PR #74, opened 2016-08-01
Use api host if schema host is undefined

mikestead:
Small update to fix fallback of host.

If `schema.host` is undefined then use host serving spec document, if one exists. [ref](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema)

Also fixed up an issue with scheme fallback where url parser returns a protocol with `:` at the end causing an invalid url

### DIFF
--- a/lib/utils/SpecManager.ts
+++ b/lib/utils/SpecManager.ts
@@ -43,7 +43,7 @@ export class SpecManager {
   /* calculate common used values */
   init() {
     let protocol;
-    const urlParts = urlParse(this._url);
+    let urlParts = urlParse(this._url);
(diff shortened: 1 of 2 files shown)


mikestead:
Not sure the issue with CI `Error: No selenium server jar found at the specified location`? Let me know if its something I need to fix up.

RomanGotsiy:
@mikestead thanks for the contribution!
Regarding CI error: everything is OK, It appeared that

> Sauce Labs addon is not available on pull request builds

I will fix this.

RomanGotsiy:
new version v1.0.1 has been just released 👌

PR #94, opened 2016-08-31
give type "string" for pipe transforms

kedashoe:
typescript was complaining about these lines

`
Argument of type 'String' is not assignable to parameter of type 'string'.
  'string' is a primitive, but 'String' is a wrapper object. Prefer using 'string' when possible.
`

telling it that these were strings made it happy


RomanGotsiy:
@kedashoe thanks!

PR #95, opened 2016-08-31
support "x-nullable" property

### LINKED ISSUE #92: Support for nullable/x-nullable
Hello, just getting started with swagger and came across ReDoc. Really great, got it working pretty quick despite having no idea what I'm doing :p

I'd like to use nullable for a couple properties in an object definition. I came across [this issue](https://github.com/OAI/OpenAPI-Specification/issues/229) and it looks like the OpenAPI folks are indeed [going to support it](https://github.com/OAI/OpenAPI-Specification/pull/741#issuecomment-235959201) in the future.

Can ReDoc add support to the ui for either/both of x-nullable/nullable? Thanks!

kedashoe:
Closes #92

For now, just supporting `x-nullable`. Once OpenAPI publishes a version with `nullable`, will be very easy to add support for that.

I just picked a light-blue-ish color for "Nullable" in the UI. If you would rather use some other color that is totally ok with me, let me know and I can update the PR or you can do it after in a separate commit.

Let me know if I didn't get to anything that should also be changed for this.

### DIFF
--- a/demo/swagger.yaml
+++ b/demo/swagger.yaml
@@ -819,6 +819,7 @@ definitions:
         type: string
         pattern: "^\\+(?:[0-9]-?){6,14}[0-9]$"
         example: +1-202-555-0192
+        x-nullable: true
       userStatus:
         description: User status
         type: integer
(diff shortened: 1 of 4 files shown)


RomanGotsiy:
@kedashoe thanks for your PR!
Will release a new minor version in a few hours :)

kedashoe:
Awesome, ty!

PR #102, opened 2016-09-06
makes basePath optional

LeFnord:
according to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object) the `basePath` isn't required,
means if they isn't provided it results in an invalid url (the `undefined` part):

bildschirmfoto 2016-09-06 um 10 50 52

this PR corrects the behaviour, so the url would be valid

bildschirmfoto 2016-09-06 um 10 51 22

### DIFF
--- a/lib/utils/SpecManager.ts
+++ b/lib/utils/SpecManager.ts
@@ -56,7 +56,8 @@ export class SpecManager {
     }
 
     let host = this._schema.host || urlParts.host;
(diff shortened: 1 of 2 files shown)


RomanGotsiy:
@LeFnord thanks!

RomanGotsiy:
I will release a new version till the end of week.

LeFnord:
👍 … cool

PR #131, opened 2016-11-01
Add option to hide hostname in method definition

bfirsh:
This allows ReDoc to be used for self-hosted tools where the hostname may not be known.


RomanGotsiy:
@bfirsh thanks for contribution!
I will publish new version during next hour

RomanGotsiy:
Something wrong with Travis :( Build failed and I can't check the reason.
So I will publish in the morning

RomanGotsiy:
@bfirsh published

PR #152, opened 2016-12-05
Add display name for tags

bfirsh:
In go-swagger, the name of the tag determines the directory that generated files end up in. Also, we want the tag names to be consistent with operationIds.

Unfortunately, these make bad names for menu items, so we want to have a way of specifying human-friendly names for the menu.


RomanGotsiy:
@bfirsh awesome PR!
Thanks.
Will be included in release till the end of the week.

RomanGotsiy:
FYI: new version `v1.6.2` has been released

bfirsh:
Sweet - thanks!

PR #175, opened 2016-12-29
Add support for second-level headings from Markdown docs

jaingaurav:


RomanGotsiy:
damn @jaingaurav you rock 🎉 

So far I see only one minor and non-trivial issue. It is a `break` on the line [146](https://github.com/Rebilly/ReDoc/pull/175/files#diff-943f2365d2291cbb72b681385e4cf31aR146):
there may exist operations that are included in more than one tag. So prepending all the parent ids to selector is the way to target element from the correct tag as methods are inside tags in HTML. With markdown headings they are not child-inside-parent so I understand why you've added this `break`.

I'm still not sure what is the best way to solve this issue though. I will post my ideas in the morning EET

jaingaurav:
@RomanGotsiy: Thanks, so I could also set the type of the menu to 'header' and only break is that is true. That might allow working with both.

RomanGotsiy:
@jaingaurav yes, the solution you suggested seems to be good 👍

jaingaurav:
Thanks @RomanGotsiy updated the diff. Seems that is was easier just checking for no metadata specified. Trying to add the `header` in `metadata.type` caused far more trouble than it was worth.

RomanGotsiy:
@jaingaurav looks great 👍 
I am going to merge this branch into master and release a new version tomorrow after adding a few tests and warnings.
Will let you know

Issue #171, opened 2016-12-21
Any plans for support xml samples for request/response ?


it-subscriptions:
If an operation has consumes and/or produces value of application/xml it would be nice to have the samples show accordingly. 

For request I can make a code sample, but ideally it would be auto generated like the JSON. 

As for responses I can't really think of any workaround.

RomanGotsiy:
@it-subscriptions there are no plans to bring auto-generated XML examples in the near future (if nobody would contribute)

But OpenAPI(Swagger) has an [`Example Object`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#example-object) using which you can provide response samples for different mime-types.
Right now only `application/json` is displayed in ReDoc but I can implement displaying `xml` sample if it's present. This should be much cheaper than implementing XML sample generation.

So in the end you will be able to provide request samples using code samples and responses using native swagger response samples .
Would it work for you?

it-subscriptions:
That would be great. Code samples for request is fine with me, so adding support for `application/xml` mime type in responses would work well as a quick solution.

Thanks!

RomanGotsiy:
Closing in favour of #492

PR #225, opened 2017-03-06
Allow JSON-like examples to be shown

brendo:
The API my company is trying to describe with Open API implements [JSON API](http://jsonapi.org/) which dictates that the Content Type be set to `application/vnd.api+json`.

We are encouraging developers to make use of the `examples` functionality for their endpoints and noticed that the current codebase only shows examples if they matched `application/json`. This essentially meant our examples were not being displayed by ReDoc.

This change does a fuzzy match to see if the content type contains anything like `json`. It's working well for us so-far, so thought we'd share back!


brendo:
I also noticed #171, so I realise there might be other ways you'd prefer to support this, and that's fine too 👍

RomanGotsiy:
Hey @brendo, thanks for PR. Sorry for long reply, a bit busy with other projects.
I will review and merge it ASAP.

brendo:
Closing in favour of #227

Issue #276, opened 2017-05-29
Issue with stick sidebar when there is content above reDoc


Alfaspider:
Hi,

I'm trying to integrate redoc into existing website but I have issue with sticky sidebar overlaping content above redoc container. I've looked trough source code and I see postion should initially be absolute but in my case it immediately changes to fixed. If I scroll down a bit, sidebar fixes itself and stays ok even if I scroll up to top again. Problem is just initial load of page.

![image](https://cloud.githubusercontent.com/assets/546762/26541343/9989995c-4455-11e7-9fab-2aa949b44f45.png)

![image](https://cloud.githubusercontent.com/assets/546762/26541378/b14c45a8-4455-11e7-82b0-539543156591.png)

RomanGotsiy:
@Alfaspider thanks for the bug report.
Will try to fix for the next version

wimleers:
We ran into this too — wonderful to see this fixed already :)

When do you think there will be a release that contains this fix?

Alfaspider:
Check out latest code and build it. Fix is already there.

RomanGotsiy:
@wimleers @Alfaspider sorry for long replies and not releasing it earlier.
New release will be ready tomorrow morning!

RomanGotsiy:
@wimleers should be fixed in the v1.16.1

PR #277, opened 2017-05-31
fix: (#276) Issue with stick sidebar when there is content above reDoc

Alfaspider:
Seems like heights are being incorrectly calculated when initialising page. I've prevented this behaviour by looking at loading class name of redoc-wrap element.


RomanGotsiy:
@Alfaspider thanks for finding!
I've updated your fix a bit. After I release a new version, could you please verify it still is working:
https://github.com/Rebilly/ReDoc/commit/4dd8a0acdff1a33057734087ac6bd534e434de03

Alfaspider:
Excellent, I didn't know about specLoading variable. Fix is now a lot cleaner.

PR #282, opened 2017-06-12
use node-sass ^4.5.3, which supports NodeJS 8

jfeltesse-mdsol:
With the current version, on macOS, trying to start the local server with NodeJS 8 results in
```
Module build failed: Error: Node Sass does not yet support your current environment: 
OS X 64-bit with Unsupported runtime (57)
```


RomanGotsiy:
@jfeltesse-mdsol thanks!

Issue #283, opened 2017-06-14
Support security OAuth2 scopes on operations


jimmyjames:
Operations may define the required OAuth2 scope. For example, in the example PetStore spec, to `POST` to the `/pet` endpoint to create a new Pet, it is required that the token have read and write permissions:

```yaml
security:
  - petstore_auth
      - 'write:pets'
      - 'read:pets'
```

You can see from the [demo](http://rebilly.github.io/ReDoc/#operation/addPet) that the required auth scopes are not shown for the operation.

Is it on the near-term roadmap to support rendering the required auth scopes for an operation? We are loving ReDoc, but this is a pretty big deal as users need to know the required auth scope.

jimmyjames:
If this is something that is agreed-upon as a good feature, I'm willing to help with a PR to get this in.

jimmyjames:
I have this working by showing any OAuth2 scopes beneath the responses for each operation (won't display if there are no oauth2 scopes for an operation):

![image](https://user-images.githubusercontent.com/276225/27339211-c92a801a-559c-11e7-94a9-507b4df0b7f5.png)

There are other UX possibilities of course, like adding the security info all in a tip (for example with a little padlock icon that shows the security scopes when hovering over, next to the operation description). 

I'll follow with a PR for what I've done if this looks like a good first step. Thanks!

Schandlich:
@jimmyjames I would love to see it be associated with the request vs. the response personally.

jimmyjames:
Yeah @Schandlich I kind of agree. What about something like this, where there's a tooltip hint floated right of the operation (similar to Swagger UI):

![image](https://user-images.githubusercontent.com/276225/27452765-7082b0a4-5759-11e7-90a7-24171c362a4e.png)

Schandlich:
@jimmyjames I think the location is right. I would use a different icon though. Excited for this!

RomanGotsiy:
@jimmyjames sorry for long reply!

This looks awesome. Proceed with PR please!

jimmyjames:
@RomanGotsiy will do, need to move it from proof-of-concept to PR-worthy (me no Angular so well 😄 ). I used a `!` since I didn't find any imgs in the project, wasn't sure if that was just because there hasn't been a need yet, or due to a design/performance consideration. I can either stick with that or try and find some type of padlock icon to use instead...

RomanGotsiy:
Start with `!` and then we will figure out how to add padlock icon in PR comments :)

[...]

PR #287, opened 2017-06-23
Addresses issue #283 by adding oauth2 security scopes to operation docs

jimmyjames:
This PR attempts to address issue discussed in issue #283 

Operations often have a specific security scope associated with them, and currently that information is not available via ReDoc. This PR:

- Renders a tooltip to the right of the operation header if there are OAuth2 security definitions for that operation.
- In order to support HTML display in the tooltip, changes dependency on [hint.css](https://github.com/chinchang/hint.css) to [html-hint.css](https://github.com/istarkov/html-hint) (based on hint.css). 

Some things to consider/review from those with more knowledge:

- Other tooltips using hint appear to work OK (probably because html-hint is based on hint), but could use some consideration from domain-experts.
- This pass only supports OAuth2 security definitions per operation as a first-pass - it does not consider api_keys. OK for a first pass?
- This implementation does not attempt to handle any `$ref` to security definitions per-operation. I don't know if this is supported/common so worth considering (perhaps at least needs to be handled for the `securityDefinitions` lookup of the descriptions?).
- Currently displays as `!` styled similar to other tips like Parameters. Might be better to display as some sort of padlock icon (like Swagger UI) to be clearer, but per issue comments starting with the `!`

### DIFF
--- a/lib/utils/spec-manager.ts
+++ b/lib/utils/spec-manager.ts
@@ -141,8 +141,8 @@ export class SpecManager {
     let that = this;
     function getSecurityDefinition(name:string, scope:string) {
       let secDefs = that._schema.securityDefinitions;
-      if (secDefs[name] && secDefs[name].type === "oauth2") {
-        let availScopes = secDefs[name].scopes
+      if (secDefs[name] && secDefs[name].type === 'oauth2') {
+        let availScopes = secDefs[name].scopes;
         if (availScopes && availScopes[scope]) {
           return availScopes[scope];
         }
@@ -162,7 +162,7 @@ export class SpecManager {
          Object.keys(scopeObj).forEach(key => {
             let val = scopeObj[key];
             let desc = getSecurityDefinition(authName, val);
-            
+
             // don't add if the security obj doesn't exist in the security definitions
             if (desc) {
               scopes.push({name: val, description: desc});
--- a/lib/utils/spec-manager.ts
+++ b/lib/utils/spec-manager.ts
@@ -137,6 +137,42 @@ export class SpecManager {
     return obj;
   }
 
+  getOperationScopes(operationPtr:string) {
+    let that = this;
+    function getSecurityDefinition(name:string, scope:string) {
+      let secDefs = that._schema.securityDefinitions;
+      if (secDefs[name] && secDefs[name].type === "oauth2") {
+        let availScopes = secDefs[name].scopes
+        if (availScopes && availScopes[scope]) {
+          return availScopes[scope];
+        }
+      }
+      return null;
+    }
+
+    let securityParams = this.byPointer(operationPtr) || [];
+    let scopes = [];
+
+    // only support one security item per operation for now
+    let base = securityParams[0];
+
+    if (base && base[Object.keys(base)[0]]) {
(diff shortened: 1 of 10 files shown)


lib/utils/spec-manager.ts
```diff
@@ -137,6 +137,42 @@ export class SpecManager {
     return obj;
   }
 
+  getOperationScopes(operationPtr:string) {
+    let that = this;
+    function getSecurityDefinition(name:string, scope:string) {
```
RomanGotsiy: Transform this function to fat-arrow function and remove `let that = this` trick:

```js
const getSecurityDefinition = (name: string, scope: string) => {
  // just use `this` here
}
```

wilsonge:
@jimmyjames are you planning on picking up the feedback changes here? I need these for a project so would be happy to pick up if you aren't planning to

jimmyjames:
Hey @wilsonge and @RomanGotsiy sorry for the late reply. @wilsonge if you have some bandwidth and can incorporate the feedback changes that would be great - it may be a little bit until I can get back to this. Thanks!

> I will test these changes on some major documentations powered by ReDoc and most likely will request some more changes.

@RomanGotsiy did you get a chance to do this?

wilsonge:
@jimmyjames See https://github.com/jimmyjames/ReDoc/pull/1

jimmyjames:
I'm not able to pull the changes and test in the next couple days; @wilsonge can you include a screenshot of how it will look? 

Given the desire to not use html-hint, I'm wondering using tooltip is not the right approach at all (unless another solution using hints that supports html content is included), and instead just include the scopes as a subsection under the operation description. It definitely makes the OAuth2 scopes more prominent in the docs instead of tucked away, for better or worse.

@RomanGotsiy curious for your feedback, as well as if you had a chance to test further.

RomanGotsiy:
Hey @jimmyjames,

sorry for delaying this for so long time. I was on vacation.
Now I'm back and I hope we will merge it soon!

Today I will release a new version of ReDoc (as there are a few fixes already merged) and then I will start testing this.
I aim to have this feature merged in the next release!

> Given the desire to not use html-hint, I'm wondering using tooltip is not the right approach at all 
Maybe we'll have to go back to `html-hint`. Will see. Give me a few days and we will merge it finally!

RomanGotsiy:
Hey @jimmyjames,

would you mind if I introduce this feature a bit later together with the rewrite to React (#327)?
This way I will have fewer things to migrate.

jimmyjames:
@RomanGotsiy sorry I missed this until now.

I understand wanting to minimize the amount of migration done. In our case, our API has many different scopes, and our most of our operations require different scopes. Not having the scopes required per operation shown in the docs is not acceptable for us, but we can fork/modify/host ourselves if we need to until migration is done and the feature is added.

RomanGotsiy:
@jimmyjames sorry for not answering for so long...
As far as you may know I'm working on React rewrite of Redoc (https://github.com/Rebilly/ReDoc/pull/357)

So basic version of this feature is already implemented there.
Hope you can help to improve it :)

Closing this PR in favor of React version.
Thanks!

Issue #286, opened 2017-06-21
Visualize deprecated operations


pgolm:
Hi

I've tried to mark some operations as deprecated like it's [documented](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) for operation objects. This option is not handled in ReDoc. Swagger UI rendered it with `text-decoration: line-through`. 

<img width="747" alt="bildschirmfoto 2017-06-21 um 11 15 35" src="https://user-images.githubusercontent.com/1444194/27376641-0a98ec86-5673-11e7-947d-180c8968728e.png">

Maybe we could add this too. What do you think?

Lytkini:
+1

avkonst:
I faced similar needs, however, I have got deprecation on per field basis, not per operation. Also, deprecation style may require rendering of a note on what the alternative / new method which replaces the deprecated stuff.

jimmyjames:
+1

I'd be fine with a simple "Warning: Deprecated" text beneath the operation header, but before the description. I think the schema object ReDoc already supports this, so should just be a matter of passing it through the controller/view and agreeing on style/text to add.

I think the suggested/new operation to use (if applicable) would need to be in the operation description itself, since the `deprecated` field is just a boolean (though .

PR #290, opened 2017-07-01
Add special rendering for deprecated operations

pgolm:
This PR adds visual hints to deprecated operations which discussed in #286.

bildschirmfoto 2017-07-01 um 11 13 05

### DIFF
--- a/demo/swagger.yaml
+++ b/demo/swagger.yaml
@@ -336,6 +336,7 @@ paths:
       summary: Finds Pets by tags
       description: 'Muliple tags can be provided with comma seperated strings. Use tag1, tag2, tag3 for testing.'
       operationId: findPetsByTags
+      deprecated: true
       produces:
         - application/xml
         - application/json
--- a/lib/components/Operation/operation.html
+++ b/lib/components/Operation/operation.html
@@ -1,13 +1,10 @@
 <div class="operation" *ngIf="operation">
   <div class="operation-content">
-    <h2 class="operation-header sharable-header">
+    <h2 class="operation-header sharable-header" [ngClass]="{deprecated: operation.deprecated}">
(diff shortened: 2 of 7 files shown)


RomanGotsiy:
Looks awesome! Thanks for kicking it off @pgolm. Sorry for long reply.

Just one thing. Could you instead of adding label `Warning: Depratecated` add `deprecated` class to operation header just same way as you did here https://github.com/Rebilly/ReDoc/pull/290/files#diff-74068f897e8aeb7897a9b86c3d9cba68R3.
Then using this class and `:after` pseudo-selector add orange badge `Deprecated` to the end of the header. + maybe make header line-crossed too.
I think it will look better. What do you think?

pgolm:
Thanks for your review. Sounds very good for me :+1:. I will change it later these day.

pgolm:
I've added the `deprecated` badge to the header. But the strikethrough header looks a little bit to heavy for me, so i left it. Any thoughts?

bildschirmfoto 2017-07-04 um 07 30 53

pgolm:
Thanks for the hint @SebastianStehle. I didn't know that. In the source i can't find any usage of it. Are there any best practice?

RomanGotsiy:
@pgolm oh my... It's so huuuge 😂 

I will merge this PR and then change the style of the badge myself if you don't mind

pgolm:
@RomanGotsiy 🙈👍

RomanGotsiy:
image

I believe it looks better this way :)

Will release a new version till the end of the week.

ssteeg-mdsol:
Thanks for adding this feature, will it be released soon?

RomanGotsiy:
Yes, will cut a new release until the end of this week!

PR #306, opened 2017-07-18
Fix event creation

wilsonge:
As per https://developer.mozilla.org/en/docs/Web/API/MouseEvent `true` isn't a valid parameter and causes an error to be thrown in typescript 2.4 (correctly). This set of parameters matches that in the catch below using `document.createEvent`

### DIFF
--- a/tests/helpers.ts
+++ b/tests/helpers.ts
@@ -21,7 +21,10 @@ export function mouseclick( element ) {
     // create a mouse click event
     var dispatchedEvent
     try {
-      dispatchedEvent = new MouseEvent('click', true);
+      dispatchedEvent = new MouseEvent('click', {
+          bubbles: true,
+          cancelable: false
+      });
     } catch (e) {
       dispatchedEvent = document.createEvent('MouseEvent');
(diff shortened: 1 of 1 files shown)


RomanGotsiy:
Hey @wilsonge,
Thanks for fixing this! 👍 
Can #305 be closed in favor of this?

wilsonge:
No - unfortunately not - the 

```
[at-loader] ./lib/services/search.service.ts:195:24 
    TS2345: Argument of type 'Schema | Schema[]' is not assignable to parameter of type 'SwaggerSchema'.
  Type 'Schema[]' has no properties in common with type 'SwaggerSchema'. 
``` 

thing isn't fixed here. and that one seems like a complete rabbit hole from a 20 minute run through it. I'm not sure I have the ability to fix that one. So this fixes one of the errors from #305 but #305 is still required for now

RomanGotsiy:
OK, thanks for diving in!
I will try to fix the second one later today.

wilsonge:
By the way the `zone.js` thing from #305 is required either way even if we fix the typescript. It looks like that's a totally broken version.

PR #313, opened 2017-07-31
In JsonFormatter, output dates as ISO 8601 strings, per OpenAPI spec.

jacobbaskin:
Without this fix, `date-time` fields in examples get output as empty objects.


RomanGotsiy:
@jacobbaskin thank you for this PR!

PR #325, opened 2017-08-16
Closes #322: Feature request: Clickable logo that points to specific url

### LINKED ISSUE #322: Feature request: Clickable logo that points to specific url
Hi!

I'd love to specify url somewhere somehow. And I'd like my logo to be clickable and point to that url.

Is it possible, now?

Thanks!

khorolets:
Summary:

* #322: Feature request: Clickable logo that points to specific url
* Add "Return to website" link to the top right that leads to `info.contacts.url` if any.
* Increase padding top for `.api-info-wrapper` when left sidebar is hiding to avoid header overlaying by top menu.

@RomanGotsiy Unfortunately I don't do Angular, so I'm not sure if `lib/components/ApiInfo/api-info.html` is correct.

I, also, added a "Return to website" button, I need it, maybe someone need it, too. (It won't show up if no `url` in `contact` section of spec is present. Also, we can hide it with css `.go-back-link { display: none }`)

And I've noticed that  `.api-info-wrapper h1` becomes overlaid by top menu on small devices (once left sidebar is hiding), so I've increased the top padding for that.

I haven't found contributors rules in the repo, so I'm not sure if it is OK to do so. So let me know if anything is wrong and needs to be edited.

Thanks. 

UPDATE

Forgot to attach screenshot of the results: 
redoc_return_link

### DIFF
--- a/lib/components/ApiInfo/api-info.scss
+++ b/lib/components/ApiInfo/api-info.scss
@@ -9,6 +9,10 @@
   @media (max-width: $right-panel-squash-breakpoint) {
     width: 100%;
   }
+
+  @media (max-width: $side-menu-mobile-breakpoint) {
+    padding-top: $section-spacing + 20px;
+  }
 }
 
 .openapi-button {
--- a/lib/components/ApiLogo/api-logo.html
+++ b/lib/components/ApiLogo/api-logo.html
@@ -1,1 +1,4 @@
-<img *ngIf="logo.imgUrl" [attr.src]="logo.imgUrl" [ngStyle]="{'background-color': logo.bgColor}">
+<a *ngIf="logo.url" href="{{logo.url}}">
+    <img *ngIf="logo.imgUrl" [attr.src]="logo.imgUrl" [ngStyle]="{'background-color': logo.bgColor}">
+</a>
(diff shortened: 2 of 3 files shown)


RomanGotsiy:
Hey @khorolets! Thanks for PR!
Great idea to use `info.contacts.url`!

I have concerns abot "return to website" button. I think it shouldn't be shown by default. I'm sure some ReDoc users don't need (and don't want) this button as they just use navigation bar on the top (for example https://apis.guru/api-doc/)
Also I don't like position of this button. Let's track this feature request separatelly. You can open another PR for this.

Regarding your padding fix, could you just put it into a separate commit with `fix:` prefix in message (it is neaded for auto changelog generation)
Thanks!

khorolets:
@RomanGotsiy ok, thanks. I will update my PR in 12 hours or so.

hlorofos:
By adding hardcoded string "return to website" we are limiting some users who want to use different language for example.

khorolets:
@hlorofos Totally agreed. Removed that change.
@RomanGotsiy Updated according to your request. Please, verify.

RomanGotsiy:
@hlorofos Great point! 👍 

@khorolets awesome! Merged! Thanks so much!!!

I will release a new release at the beginning of the next week

khorolets:
@RomanGotsiy Thank you!

Issue #98, opened 2016-09-02
Feature Request: Navigation Sorting


gmr:
I just saw this tonight and have already started to roll it out for my APIs.

One nit is the ordering of the sidebar navigation seems to be fairly random.

It'd be awesome if I could sort endpoints in the navigation with an `x-` variable or by setting a configuration value that told it to sort alpha by `operationId` or `summary`.

RomanGotsiy:
> already started to roll out
👍  🎉 

Currently, methods are grouped by tags (similar to the SwaggerUI) and tags (top level menu items) are sorted according to the order of [`tags` field](https://github.com/OAI/OpenAPI-Specification/blob/master//versions/2.0.md#swagger-object).

Inside tags methods are ordered in the same order as the appear in json/yaml spec (yes I know json doesn't have an order but in all modern javascript engines it appears to have 😄 )

If you have many endpoints - tags are a good way to organize things.
If this doesn't work for you, we can discuss some option or `x-` vendor extension.

gmr:
I have the tags for organizing, I was more interested in the actual endpoint/method sorting.  I've also noticed request and response parameter sorting doesn't match the spec as well.

It seems the sorting is deterministic, but not based upon the document as far as I can tell.

I was thinking an ``x-sort-order`` or something might be a good way to go, but perhaps I'm missing something.

Thanks BTW, this is exactly what I was looking for when I started [this](https://github.com/gmr/oaui/tree/initial-version).

jonyw4:
I need this feature too. How I can support? +1

slavcodev:
I'm interesting in the order of side menu too. And this is my proposal.

The headers from the `info` description are rendering as menu items now. Also ReDoc already supports kind of "widgets system" using HTML comment for the injection of the security definitions. I propose to use these features further. There is example how to inject operations in the `info` description:

~~~yaml
info:
  description: |
    # Introduction
    ...
    # Group e.g. Pet

    ## Add a new pet to the store
    <!-- see: #/paths/~1pet/post -->

    ## Update an existing pet
    <!-- see: #/paths/~1pet/put -->
tags:
  - name: Pet
    description: Everything about your Pets
  - name: Store
    description: Access to Petstore orders
paths:
  /pet:
    post:
      tags:
        - pet
      summary: Add a new pet to the store
    put:
      tags:
        - pet
      summary: Update an existing pet
~~~

The conventional rules:

1) The operation summary should be same as the header title.
[...]

PR #328, opened 2017-08-17
Feature/sorting

Darkey999:
As requested in: #98 I've added sorting:

- By default, it first sorts elements by the name of the operation (delete, get, post etc.), then by the name of the endpoint (add pet, update pet etc.)
- You can use `x-sort-order` to override default sorting and arrange endpoints in any way you want, check redoc-vendor-extensions.md for description and example

If you want to see the effect, click [here](http://maksok.cba.pl/api/testing.html).
To get the file, click [here](https://darkey999.github.io/redoc-sort.min.js).

### DIFF
--- a/docs/redoc-vendor-extensions.md
+++ b/docs/redoc-vendor-extensions.md
@@ -243,3 +243,40 @@ PayPalPayment:
 
 In the example above the names of definitions (`PayPalPayment`) are named differently than
 names in the payload (`paypal`) which is not supported by default `discriminator`.
+
+### Sort endpoints vendor extensions
+#### x-sort-order
+| Field Name     |	Type	  | Description |
+| :------------- | :------: | :---------- |
+| x-sort-order | int | specifies sorting order |
+
+###### Usage in ReDoc
+By using `x-sort-order` you can override default alphabetical sorting and arrange elements in any way you want.
+
+
+Add `x-sort-order` with number from 0 to number of elements in group - 1 (e.g. for 3 items, the scope is 0-2).
+Endpoints with lower `x-sort-order` number come first in a group.
+
+###### x-sort-order example
+
+```
+/store/inventory:
+  get:
+    tags:
+      - store
+    summary: Return pet inventories by status
+    x-sort-order: 1
+    description: Returns a map of status codes to quantities
+      
+/store/order:
+  post:
+    tags:
+      - store
+    summary: Place an order for a pet
+    x-sort-order: 0
+    description: Places an order for a pet
+
+```
+
+In this example "Place an order for a pet" endpoint will come before "Return pet inventories by status"
+(because of the `x-sort-order` property).
--- a/lib/components/SideMenu/fixed-array.ts
+++ b/lib/components/SideMenu/fixed-array.ts
@@ -1,0 +1,188 @@
+export let fixedMenuItemsList =
+[
+    {
+        name: "Pagination",
+        id: "section/Pagination",
+        items: []
+    },
+    {
+        name: "JSONP",
+        id: "section/JSONP",
+        items: []
+    },
+    {
+        name: "Authentication",
+        id: "section/Authentication",
+        items: []
+    },
+    {
+        name: "pet",
+        id: "tag/pet",
+        items: [
+            {
+            	name: "Deletes a pet",
+            	id: "/paths/~1pet~1{petId}/delete",
+            	metadata: {operation: "delete"}
+            },
+            {
+            	name: "Find pet by ID",
+            	id: "/paths/~1pet~1{petId}/get",
+            	metadata: {operation: "get"}
+            },
+            {
+            	name: "Finds Pets by status",
+            	id: "/paths/~1pet~1findByStatus/get",
+            	metadata: {operation: "get"}
+            },
+            {
+            	name: "Finds Pets by tags",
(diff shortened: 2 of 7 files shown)


RomanGotsiy:
Hey @Darkey999!

Thanks for you PR!
I will review it on Monday as I am not here tomorrow and on the weekends

docs/redoc-vendor-extensions.md
```diff
@@ -243,3 +243,40 @@ PayPalPayment:
 
 In the example above the names of definitions (`PayPalPayment`) are named differently than
 names in the payload (`paypal`) which is not supported by default `discriminator`.
+
+### Sort endpoints vendor extensions
```
RomanGotsiy: `x-sort-order` should be placed under `Operation Object vendor extensions` not to it's own section

lib/components/SideMenu/fixed-array.ts
```diff
@@ -0,0 +1,188 @@
+export let fixedMenuItemsList =
```
RomanGotsiy: rename this file so it's clear that it is used in tests, e.g. `fixed-array.mock.ts`

lib/components/SideMenu/side-menu.ts
```diff
@@ -113,7 +114,21 @@ export class SideMenu implements OnInit, OnDestroy {
 
   init() {
     this.menuItems = this.menuService.items;
-
+    let firstMenuItemsLength = this.menuItems.length;
+
+    for (var firstMenuItemIndex = 0; firstMenuItemIndex < firstMenuItemsLength; firstMenuItemIndex++) {
```
RomanGotsiy: Menu items can have more than 2 levels. Better to not hardcode two levels here but replace this with a recursive sorting algorithm which will support any nesting

Also line 124 may be executed multiple for each `firstMenuItemIndex` which is unnecessary work

lib/components/SideMenu/sort.ts
```diff
@@ -0,0 +1,23 @@
+export const endpointComparator = (a, b) => {
```
RomanGotsiy: Could you change this method to sort items **if only** `sortOrder` is specified?
Otherwise, this will introduce breaking changes as many ReDoc users rely on the current order (preserved from JSON)

lib/services/menu.service.ts
```diff
@@ -332,6 +332,15 @@ export class MenuService {
 
     let res = [];
     for (let operationInfo of tag.operations) {
+      let xSortOrder;
+      let orderInfo = operationInfo['x-sort-order'];
+
+      if (orderInfo != null) {
```
RomanGotsiy: why we need all these checks here? I would just replace all this with: `xSortOrder = operationInfo['x-sort-order'];`

lib/services/menu.service.ts
```diff
@@ -341,7 +350,8 @@ export class MenuService {
           pointer: operationInfo._pointer,
           operationId: operationInfo.operationId,
           operation: operationInfo.operation,
-          deprecated: !!operationInfo.deprecated
+          deprecated: !!operationInfo.deprecated,
+          sortOrder: xSortOrder+""
```
RomanGotsiy: Do not transform this to string. This way "11" and "9" will be sorted incorrectly as "9" is greater that "11" in terms of string comparison:

```js
"9".localeCompare("11") === 1
"12".localeCompare("11") === 1
```

as you can see sort order would be broken

Darkey999:
Hey @RomanGotsiy!

Thanks for checking my PR, I'll make all necessary changes as fast as I can.

CumpsD:
No idea if there are still plans for this, but I just used this code to get my operations sorted.

`sort.ts` just for operation comparing:
```
export const endpointComparator = (a, b) => {
  return (a.metadata.operation != null && b.metadata.operation != null)
    ? a.metadata.operation.localeCompare(b.metadata.operation)
    : 0;
}
```

`side-menu.ts` has a recursive menu sorter added:
```
  sortMenuItems(menuItems: MenuItem[]) {
    let menuItemsLength = menuItems.length;
    for (var menuItemIndex = 0; menuItemIndex < menuItemsLength; menuItemIndex++) {
      let subItems = menuItems[menuItemIndex].items;
      if (subItems !== undefined && subItems !== null && subItems.length > 0) {
        subItems = this.sortMenuItems(subItems);
      }
    }

    menuItems.sort(endpointComparator);
    return menuItems;
  }
```

and `init` has: `this.menuItems = this.sortMenuItems(this.menuService.items);`

Darkey999:
Thank you, I want to finish all necessary changes, but since my academic year started, I can't find time for this. Bus as I said, I'll do it.

RomanGotsiy:
If anyone willing to rebase this PR on the [`react-rewrite`](https://github.com/Rebilly/ReDoc/tree/react-rewrite) branch it would be just awesome!

jsamr:
@RomanGotsiy @Darkey999 Any updates ?

RomanGotsiy:
@jsamr this PR is opened against 1.x version.
It needs to be rewritten against 2.x.

I don't have much time to do it right now.
Any help would be appreciated!

RomanHotsiy:
1.x version of Redoc is in maintenance mode. New features won't be added.

Closing this PR. If somebody want's this feature, please open a new PR against the React version.

Thanks @Darkey999 for amazing job and sorry for not merging this.

PR #334, opened 2017-08-29
add example to property infos

SiCoe:
As well as the _Default:_ being shown for each property. also allow the _Example:_ to be shown. Much like they are for parameters.

### DIFF
--- a/lib/components/JsonSchema/json-schema.html
+++ b/lib/components/JsonSchema/json-schema.html
@@ -77,6 +77,9 @@
               <div class="param-default" *ngIf="prop.default != null">
                 <span class="param-default-value">{{prop.default | json}}</span>
               </div>
+              <div class="param-example" *ngIf="prop.example != null">
+                <span class="param-example-value">{{prop.example | json}}</span>
+              </div>
               <div *ngIf="prop.enum && !prop.isDiscriminator" class="param-enum">
                 <span *ngFor="let enumItem of prop.enum" class="param-enum-value {{enumItem.type}}"> {{enumItem.val | json}} </span>
               </div>


RomanGotsiy:
Hi @SiCoe, I intentionally didn't add `Example` to property infos as they all are used in the right panel samples.

This will be duplicated information if we add examples to each property:

image

So I think we should not add examples to properties. Why do you need them? What is the use case?

SiCoe:
Hi @RomanGotsiy, thanks for reply.

My first use case was for a **percent** property info within a complex type.
Setting `Example: 85` (opposed to `Example: 0.85`) easily showed how the percentage should be represented.

I have to admit, I only noticed the request sample on the right had the example value in it after I'd already sent you the PR.

I agree that it's a duplicate of information, but I think the reason I didn't see it was due to the **percent** property info being nested a couple layers down the complex object: while I expanded the middle panel to get to the property info, the right panel's request sample was still collapsed and therefore the example value was hidden.

RomanGotsiy:
I think this specific info may be just included into the description or into the combination of `minimum` and `maximum`, as ReDoc renders this info (see `[1..5]`):

image

I have concerns whether to add this by default because a few ReDoc users heavily use `example` in each field and I'm not sure if their docs will look OK if we enable this by default in the middle panel.

If you are not satisfied with the workarounds we can think of disabling/enabling this feature by some ReDoc option

SiCoe:
I understand, thank you for your time.

I think we'll stick with using the _request samples_ as you say.

If it becomes apparent that an example next to the property info will be of significant benefit then I'll let you know and we can look at having it as a configuration option.

RomanGotsiy:
Awesome!
Thank you for your understanding!

Issue #345, opened 2017-09-21
Add x-permissions feature


zucher:
Some api are accessible one for some user with a specific level of permission: superadmin admin, user, guest,...

I propose an x-permissions array object:
{
"name": "string",
"summary": "string",
"description": string"
}

example:
"x-permissions": [{
                    "name": "superadmin",
                    "title": "Solution Admin",
                    "description": "Solution Admin has high privileges on Pet API "
                }...]

With this information associated to operation api, a simple tag with mouse over or everything else is specified below the Operation label. like

## Get Pets operation
*Permissions* `Solution Admin` `User`...

The description could be either detailed below the authentication chapter or into every operation (a little bit heavy)

We also may add x-permissions list into swagger top level like tags, just refer to them into operation

zucher:
See http://apidocjs.com/#param-api-permission for sample declaration into an other kind of rest api declaration

zucher:
I do the job this the second solution, like tags after header description with under operation the lable for each of them
![image](https://user-images.githubusercontent.com/3034596/30899838-e2e75056-a361-11e7-9b7a-298ddefbe210.png)

zucher:
I have to add documentation and tests

zucher:
See https://github.com/Rebilly/ReDoc/pull/349

regards

RomanGotsiy:
Closing this issue.
Reasons explained in the corresponding PR.

Thanks!

PR #349, opened 2017-09-27
Add support of x-permissions vendor extension

zucher:
according to Add x-permissions feature #345, find the feature implementation


jcarres-mdsol:
I guess now this can't go in till the react rewrite is finished but something like this would be welcome in the future? @RomanGotsiy

RomanGotsiy:
@zucher sorry for ignoring this PR for sooooo long time 🙈

This feature is too specific to your use case and we will not put it into ReDoc core.

I am going to close this PR but you are free to maintain your own fork though.
If you don't want to maintain your own fork I would recommend you to switch to a more generic solution based on vendor extensions proposed by @ingoclaro in https://github.com/Rebilly/ReDoc/pull/552 once it get merged.

Thanks for understanding!

Issue #113, opened 2016-09-16
Better error messages for diagnosis


natke:
When I try to render the documentation for my swagger.json spec, I get the following error:

Oops... ReDoc failed to render this spec

Unable to resolve $ref pointer 

It's a rather large specification and probably has an error in it, but I need some more information to locate the error (if there is one). Is it possible to get access to some more information about the error?

I tried to run the demo locally but could not run the following step:

npm run jspm-install

npm ERR! Darwin 15.4.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "jspm-install"
npm ERR! node v4.4.3
npm ERR! npm  v2.15.1

npm ERR! missing script: jspm-install

RomanGotsiy:
@natke, yes this error message definitely should be improved.
I will check this.

Actually, this error message comes from tool I use for reference resolving so most likely your spec has some unresolvable references. Try to validate your spec using `editor.swagger.io`.

Regarding `npm run jspm-install`: it's not actual anymore, I just forgot to update steps. Thanks for reporting this.

natke:
Thanks for the quick reply.

Is the tool you talking about: https://www.npmjs.com/package/json-schema-ref-parser?

I ran the spec through the tool manually and it generated the resolved spec fine. But maybe I didn't exactly replicate the calls to this tool in the way that redoc is doing? Is it possible to summarize the API calls you are using?

RomanGotsiy:
@natke, would be great if you're able to provide me with a minimal reproducible spec or the actual spec you are trying to visualize with ReDoc?

natke:
Hi again, I did manage to get it working. I think the problem was that I was trying to serve a file. Once I served both my index.html and my spec over http it worked fine.

I think that error message is not actually from the parser - it is from your code, but it looks like there is an unresolved promise somewhere in the call chain. Might be worth looking into, and adding some better diagnostics and tests.

RomanGotsiy:
@natke, If you don't mind I will keep this issue open as it is not fixed: error messages sometimes are not clear

natke:
No problem - just wanted to indicate that I didn't have any further issues

krsyoung:
@natke curious if I'm getting the same thing you were.  What I was trying to do was run redoc with a local index.html and swagger file (i.e. just open them from the filesystem vs serving via HTTP).  Is this what you were originally doing when you received the error?

In my case I get a full error as: 
```
[...]

PR #380, opened 2017-11-20
switch from json-schema-ref-parser's bundle method to dereference

vegyraupe:
Fix / work-around for [Issue 311](https://github.com/Rebilly/ReDoc/issues/311)

### DIFF
--- a/lib/utils/spec-manager.ts
+++ b/lib/utils/spec-manager.ts
@@ -44,7 +44,7 @@ export class SpecManager {
   load(urlOrObject: string|Object) {
     let promise = new Promise((resolve, reject) => {
       this.parser = new JsonSchemaRefParser();
-      this.parser.bundle(urlOrObject, {http: {withCredentials: false}})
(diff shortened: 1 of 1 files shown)


RomanGotsiy:
@vegyraupe thanks for this PR.
But it makes unit tests fail. I will investigate though.
But, could you please verify if the issue #113 still occurs on the 2.0.0-alpha.1.
It is complete ReDoc rewrite using React so the issue may be fixed now.

Thanks!

vegyraupe:
@RomanGotsiy I tried with alpha.5 and the error still occurs. Switching bundle to dereference works again. The problem occurs when the same external reference is being used twice. This results in bundle to only include it once and returning a relative reference instead of the dereferenced object.

RomanGotsiy:
@vegyraupe could you share your spec? or at least minimal reproducible sample?

RomanGotsiy:
Closing this as stale

PR #386, opened 2017-11-28
Add note that x-logo SHOULD use absolute URLs

MikeRalphson:
Have seen a number of definitions designed for use with ReDoc using a relative logo URL. This 'breaks' when someone else (e.g. APIs.guru) hosts the API definition.


RomanGotsiy:
makes sense! Thanks @MikeRalphson !

PR #396, opened 2017-12-20
Make a single request for Google Fonts

brendo:
Just noticed there were two HTTP requests when we can do it one :)


RomanGotsiy:
@brendo nice catch! 👍

RomanGotsiy:
I will release a new package tomorrow.

brendo:
Sounds great, thank you!

PR #403, opened 2018-01-10
Update README.md

ilMattion:
Added one simple command: npm run build:module.

On clean branch, if you don't execute this command you have not "compiled" and application raise exception.


RomanGotsiy:
Hey @ilMattion, The message you were getting is not an error, it's just a warning.
There is a better fix for this: I just silenced this with new typescript feature (https://github.com/Rebilly/ReDoc/commit/5925431dbf3c45a4096b818cfa45b5ff6a8dc114)
So there is no need to update README anymore! Anyway thanks for the help!

PR #407, opened 2018-01-12
Use safer mark.js version (npm)

julmot:


julmot:
Failing test doesn't appear to be cause by my change: `Cannot start ChromeHeadless`

RomanGotsiy:
Thanks @julmot for this PR and for awesome work with mark.js !

PR #420, opened 2018-02-22
fix the media queries utils so it gets the values from the current theme

CesarLanderos:
## Description

the media queries helpers use the base theme breakpoints instead of the current theme ones set on the standalone component props, this PR fixes that.

### DIFF
--- a/src/styled-components.ts
+++ b/src/styled-components.ts
@@ -1,6 +1,6 @@
 import * as styledComponents from 'styled-components';
 
-import theme, { ThemeInterface } from './theme';
+import { ThemeInterface } from './theme';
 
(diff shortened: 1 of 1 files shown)


CesarLanderos:
seems like i need to fix the tests so the theme is injected in the context 😞

RomanGotsiy:
@CesarLanderos to fix this change `args[0]` to `props => props.`, e.g.:

```css
@media (max-width: ${args[0].theme.breakpoints[breakpoint]}) {
```
should become

```css
@media (max-width: ${props => props.theme.breakpoints[breakpoint]}) {
```

CesarLanderos:
\o/

done 😃

RomanGotsiy:
@CesarLanderos thanks! 👍

PR #436, opened 2018-03-14
React Rewrite: A couple minor bug fixes

MrBlenny:
* SelectOnClick has been fixed. This was crashing.
* When OpenAPIParser was dereferencing a schema and the ref was not found, it would return undefined. This would crash other segments of the code such as `if (schema.allOf === undefined) {`

### DIFF
--- a/demo/index.tsx
+++ b/demo/index.tsx
@@ -24,7 +24,7 @@ const DEFAULT_SPEC = 'openapi.yaml';
 class DemoApp extends React.Component<
   {},
   { specUrl: string; dropdownOpen: boolean; cors: boolean }
-  > {
+> {
   constructor(props) {
     super(props);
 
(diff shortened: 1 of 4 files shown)


demo/index.tsx
```diff
@@ -24,7 +24,7 @@ const DEFAULT_SPEC = 'openapi.yaml';
 class DemoApp extends React.Component<
   {},
   { specUrl: string; dropdownOpen: boolean; cors: boolean }
-> {
+  > {
```
RomanGotsiy: could you please remove space here

MrBlenny: Done

RomanGotsiy:
Nice 👍
Thanks again!

RomanGotsiy:
fyi, you still don't appear on Contributors page but you will once I merge this branch into master. GitHub doesn't count contributions to non-default branches.

MrBlenny:
All good. Would be wonderful if you could publish another alpha to npm :)

RomanGotsiy:
released

PR #441, opened 2018-03-23
Modify the peerDependencies to reflect the need for react 16.3.0

fritz-c:
When using the alpha build of ReDoc, running in an environment with React 16.2.0 results in an error due to the missing `createContext` API. This PR updates the package.json `peerDependencies` to reflect the need for react >= 16.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,8 +108,8 @@
     "yaml-js": "^0.2.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


RomanGotsiy:
Seems that something wrong is with Travis and env variables. Anyway looks good!
Thanks for PR!

PR #511, opened 2018-05-28
MergeAllOf takes items into account

jderusse:
This PR allows
```
      allOf:
        - type: array
          items:
            type: object
            properties:
              id:
                type: string
        - type: array
          items:
            type: object
            properties:
              name:
                type: string

```
And render it as

```
      type: array
      items:
        type: object
        properties:
          id:
            type: string
          name:
            type: string
```

### DIFF
--- a/src/services/OpenAPIParser.ts
+++ b/src/services/OpenAPIParser.ts
@@ -233,6 +233,15 @@ export class OpenAPIParser {
         }
       }
 
+      if (subSchema.items !== undefined) {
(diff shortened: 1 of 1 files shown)


jderusse:
Looks like this is not the right way to fix the circual issue.

It goes beyond my understanding of the code. I'm closing this PR unless someone give me hints

RomanGotsiy:
LGTM! Nice job 👍 
Thanks @jderusse

PR #516, opened 2018-05-29
Return 1 as exit code if an error happens in the cli

### LINKED ISSUE #515: redoc-cli process always returns with exit code 0
The command line interface always returns the exit code `0`, even if an error occurs. I use the cli from a Maven build which then does not stop. But the build should fail as soon as an error occurs.

brushmate:
Fixes #515


RomanGotsiy:
@brushmate nice 👍 
Thanks!

PR #518, opened 2018-05-29
Run Yargs in strict mode and return exit code 1 if the command is invalid

### LINKED ISSUE #517: CLI just exits normally if an invalid command is called
I just forgot to add the `bundle` sub-command to my `redoc-cli` and it always returned with exit code `0`. I was almost freaking out because I didn't notice that I missed the sub-command. Thus it would be better to fail if an invalid sub-command was sopecified.

brushmate:
Fixes #517

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -32,7 +32,7 @@ const BUNDLES_DIR = dirname(require.resolve('redoc'));
 
 /* tslint:disable-next-line */
 YargsParser.command(
-  'serve [spec]',
+  'serve <spec>',
   'start the server',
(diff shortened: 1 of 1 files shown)


brushmate:
After reading more on Yargs I found out that positional parameters in the form of `[param]` are optional. To make them required they must look like `<param>`.

With the latest changes in this PR usage will be printed if the program was not called correctly. If an error happens in the business logic the stack is printed but not the usage. Exit code will be 1 in any case.

RomanGotsiy:
Looks awesome! thank you!

PR #519, opened 2018-05-30
Watch the directory that contains the specified spec instead of the spec itself

### LINKED ISSUE #514: redoc-cli - Add Option to --watch Directory instead of a single File
`redoc-cli serve openapi.yaml --watch` seems to only watch for changes in `openapi.yaml`.

As I have my docs in different files, it would be nice to be able to watch a whole directory.

Small example:
```yaml
openapi: '3.0.0'

paths:
  $ref: './path.yaml'
```

brushmate:
Fixes #514

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -108,7 +108,7 @@ YargsParser.command(
         redocOptions: argv.options || {},
       });
     },
-  )
+)
   .demandCommand()
   .options('t', {
     alias: 'template',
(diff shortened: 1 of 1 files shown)


cli/index.ts
```diff
@@ -159,11 +159,17 @@ async function serve(port: number, pathToSpec: string, options: Options = {}) {
...
+    };
+
     watch(
-      pathToSpec,
+      pathToSpecDirectory,
+      watchOptions,
       debounce(async (event, filename) => {
-        if (event === 'change' || (event === 'rename' && existsSync(filename))) {
-          console.log(`${pathToSpec} changed, updating docs`);
+        if (event === 'change' || event === 'rename') {
```
brushmate: I am using IntelliJ and it has a feature called 'safe write'. It first renames the original file to `<original_name>___jb_old___`, writes the new content to the original file and then deletes the copy. I think this happens so fast that `existsSync(filename)` always returns `false`. Therefore I removed the check.

RomanGotsiy: I am not sure about this change. I will test it locally with my editor on weekends and then will merge it!
thanks!

RomanGotsiy:
Thanks, @brushmate!
👍

PR #526, opened 2018-06-05
Use item's title in array definition

jderusse:
Given the following schema
```
FooCollection:
  type: array
  items:
    type: object
    title: Foo
    properties:
      bar:
        type: string
```
This PR renders `Array of object (Foo)`

### DIFF
--- a/src/services/models/Schema.ts
+++ b/src/services/models/Schema.ts
@@ -30,6 +30,7 @@ export class SchemaModel {
   isCircular: boolean = false;
 
   format?: string;
+  displayFormat?: string;
   nullable: boolean;
   deprecated: boolean;
   pattern?: string;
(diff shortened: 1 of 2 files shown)


src/services/models/Schema.ts
```diff
@@ -139,6 +139,8 @@ export class SchemaModel {
       this.items = new SchemaModel(parser, schema.items, this._$ref + '/items', this.options);
       this.displayType = this.items.displayType;
       this.typePrefix = this.items.typePrefix + 'Array of ';
+      this.format = this.items.format;
```
RomanGotsiy: Could you remove this line? Schema model is used not only for rendering. It may be used for other purposes so copying `format` field from `items` is not quite valid.

The correct approach would be to introduce `displayFormat` field (similar to `displayType`) which is used only for rendering purposes.

jderusse: done. Do you think the same "rule" should be applied for the `title` field?

RomanGotsiy: Nice! Thanks!

> Do you think the same "rule" should be applied for the title field?

I don't think so.

PR #534, opened 2018-06-13
parse discriminator value from the referenced scheme

levpachmanov:
Add vendor extension used by swagger-codegen for Swagger 2.0 as an alternative to OpenAPI 3.0 mapping feature.
For example: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/model.mustache


RomanGotsiy:
Thanks for PR 👍

PR #539, opened 2018-06-15
fix mergeAllOf with readOnly/writeOnly

frodeaa:
resets the readOnly/writeOnly to default (undefined)
before merging it with next subschema

ref: https://github.com/Rebilly/ReDoc/issues/532
rel: https://github.com/Rebilly/ReDoc/issues/422

### DIFF
--- a/src/services/OpenAPIParser.ts
+++ b/src/services/OpenAPIParser.ts
@@ -225,6 +225,10 @@ export class OpenAPIParser {
             receiver.properties[prop] = subSchema.properties[prop];
           } else {
             // merge inner properties
+            // let readOnly/writeOnly be overridden if property is redefined
+            // i.e. fallback on default value of readOnly/writeOnly when merging
+            delete receiver.properties[prop].readOnly;
+            delete receiver.properties[prop].writeOnly;
             receiver.properties[prop] = this.mergeAllOf(
               { allOf: [receiver.properties[prop], subSchema.properties[prop]] },
               $ref + '/properties/' + prop,
--- a/src/services/__tests__/OpenAPIParser.test.ts
+++ b/src/services/__tests__/OpenAPIParser.test.ts
@@ -1,0 +1,109 @@
+import { convertObj } from 'swagger2openapi';
+import { OpenAPIParser, RedocNormalizedOptions } from '../';
+
+const options = new RedocNormalizedOptions({});
+describe('OpenAPIParser mergeAllOf', async () => {
+  const parser = new OpenAPIParser(
+    { openapi: '3.0', info: { title: 'test', version: '0' }, paths: {} },
+    undefined,
+    options,
+  );
+
+  it('merge readOnly from allOf', () => {
+    expect(parser.mergeAllOf({
+      allOf: [
+        {
+          properties: {
+            id: { type: 'string', readOnly: true },
+            name: { type: 'string' },
+          },
+        },
+        { properties: { name: { type: 'string', readOnly: true } } },
+      ],
+    }, '', false).properties).toEqual({
+      id: {
+        type: 'string',
+        readOnly: true,
+      },
+      name: {
+        type: 'string',
+        allOf: undefined,
+        parentRefs: [],
+        readOnly: true,
+      },
+    });
+  });
+
+  it('merge writeOnly from allOf', () => {
+    expect(parser.mergeAllOf({
+      allOf: [
+        {
+          properties: {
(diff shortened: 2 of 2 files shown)


RomanGotsiy:
Thanks for PR but I don't understand why you consider such behaviour valid.

Taking the schema bellow I would expect `id` field to be readOnly in the resulting view.

```js
{
  allOf: [
    { properties: { id: { type: 'string' } }, readOnly: true },
    { properties: { id: { type: 'string' } } },
  ];
}
```

Omitting a keyword (`readOnly` in our case but valid for any property) in the second `allOf` subschema doesn't negate corresponding from the previous.

I think you can achieve the expected behaviour using the following approach:

```js
{
  allOf: [
    { properties: { id: { type: 'string' } }, readOnly: true },
    { properties: { id: { type: 'string' } }, readOnly: false },
  ];
}
```

Do I understand it correctly?

frodeaa:
Possibly, but the change was suggested to keep the behavior from v1.x, where you did not need to set readOnly=true when overwriting a propert. Will check it out

RomanGotsiy:
Properties in v1.x were not merged at all. They were overwritten completely. There was a TODO note to finish this: https://github.com/Rebilly/ReDoc/blob/v1.x/lib/services/schema-normalizer.service.ts#L168 :)

So, redoc@1.x behaviour was not completely correct

Does it make sense now?

frodeaa:
yes that explains the behaviour of  redoc@1.x,

But I'm not sure the current implementation of the merge given the spec for [readOnly/writeOnly](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-20) is correct. Both readOnly/writeOnly has default value set to `false`, so setting readOnly/writeOnly to `false` to overwrite it should not be necessary.

```
{
  allOf: [
    { properties: { id: { type: 'string' } }, readOnly: true },
    { properties: { id: { type: 'string' } }, readOnly: false },
  ];
}
is equal to:
{
  allOf: [
    { properties: { id: { type: 'string' } }, readOnly: true },
    { properties: { id: { type: 'string' } } },
  ];
}
```

setting readOnly to `false` is however a solution that solves my issue. 

It was a surprise that the merge order is the reverse of the `allOf` ordering, but that is fine... Thanks 

```yml
          schema:
            type: object
            allOf:
              - properties:
                  id:
                    type: string
                    readOnly: false
              - $ref: '#/definitions/ObjectWithReadOnlyId'
```

RomanGotsiy:
> It was a surprise that the merge order is the reverse of the allOf ordering, but that is fine
that surprises me too 🙈 I'll check. Thanks for pointing it out!

> But I'm not sure the current implementation of the merge given the spec for readOnly/writeOnly is correct. 

actually, both your and mine implementations are wrong TBH according to spec. Here is short explanation why:

`allOf` is not merge, its semantic is taken directly from JSONSchema:

>  An instance validates successfully against this keyword if it validates successfully against all schemas defined by this keyword's value.

So we have undefined behaviour here as `readOnly` has nothing to validation without the context.
But if we add context by assuming that our instance (payload) is for writing and let's use the following spec:
```yaml
schema:
  type: object
  allOf:
    - properties:
        id:
          type: string
          readOnly: true
    - properties:
        id:
          type: string
```

Now, the following payload

```js
{ id: 'test' }
```

is valid against  the first subschema (without `readOnly` which means `false` by default) but is invalid against the second (it has `id` field which is `readOnly` and we know we do write) so it is invalid against the whole spec because of `allOf` semantic. So second subschema doesn't negate the `readOnly` from the first one and won't negate even if we explicitly add `readOnly: false`. Morover, order doesn't matter here.

So technically if you have `readOnly: true` specified somewhere it is impossible to overwrite it with `allOf`. Does it make sense to you?

The correct solution for you is to remove `readOnly` from the `ObjectWithReadOnlyId` definition and have separate definitions for `Object` and `ObjectWithReadOnlyId` where `ObjectWithReadOnlyId` is specified as bellow:

```yaml
ObjectWithReadOnlyId:
  type: object
  allOf:
    - properties:
        id:
          readOnly: true
    - $ref: '#/definitions/Object'
```

I may fix the current wrong ReDoc's behaviour in the future so I recomend you to fix your spec as above.

All this complex behaviour is caused by complexity of json schema and openapi. Most people don't understand `allOf` correctly and think of it as of simple merge which is not true.

frodeaa:
ah, that makes sense

Issue #522, opened 2018-05-31
support to show extension labels


ingoclaro:
would be good to have an easy way to render the extension information that the different entities have (the x-* extensions).

For example swagger-ui shows them if you pass `?showExtensions=true`, like this:
![swagger_ui](https://user-images.githubusercontent.com/1280277/40799567-d1ed8784-64c2-11e8-8fa8-e254bc7666bf.jpg)

For ReDoc I was thinking on having a configuration option and be able to either pass a list of which extensions to render, maybe with an `all` option that shows them all; or just a simple setting to only render the ones not used by ReDoc. eg:

```
config: {
  renderExtensions: true
}
```
It could be rendered under the description:

![redoc](https://user-images.githubusercontent.com/1280277/40799938-ce0554de-64c3-11e8-906b-098e06bc50ce.jpg)

This makes it simple to show that information.

---

A much more involved feature could be to pass in a react component to handle the rendering of the extensions, like this:
```
options = {
  extensionComponent: MyExtensionComponent
}
```
which could receive as props the node and the extension name. The component could delegate the rendering to a different component if needed or just return null if you don't want to be rendered.

In any case, I think this is far more advanced than what I actually need, but wanted to mention it as some food for thought that would allow for a rich customization of ReDoc.

ingoclaro:
by the way, if you like the idea I could implement this. I want to agree on the approach before coding.

zucher:
Hi @ingoclaro, 

Your Extension solution seems fine.
I'm not fan of the query activation method, but the usage of javascript embedded options activation looks better.

ingoclaro:
The query param is how swagger-ui does it

RomanGotsiy:
Hey @ingoclaro,

thanks for your feature request. It looks reasonable.

> I was thinking on having a configuration option and be able to either pass a list of which extensions to render, maybe with an all option that shows them all

I think it will be a good start.

The one thing I am not sure about is the location of the information. E.g. for your docs it's logical to have vendor extensions rendered below the field description/operation name but for someone's else it may be opposite... We can't satisfy everyone for sure though.

PRs are welcome

lornajane:
This has been inactive and I think it will be difficult to generalise about what the right behaviour is for everyone, so I'm closing it as something we could revisit after the next release.

PR #552, opened 2018-06-28
add the option to render vendor extensions

ingoclaro:
implements #522

### DIFF
--- a/src/components/Fields/Extensions.tsx
+++ b/src/components/Fields/Extensions.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import { isRedocExtension } from '../../utils/openapi';
 import { OptionsContext } from '../OptionsProvider';
 
 import { SchemaModel } from '../../services/models';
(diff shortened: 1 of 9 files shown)


src/components/Fields/Extensions.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { SchemaModel } from '../../services/models';
+import { FieldDetail } from './FieldDetail';
+
+export interface ExtensionsProps {
+  schema: SchemaModel;
+}
+
+export class Extensions extends React.PureComponent<ExtensionsProps> {
+  redocExtensions = [
```
RomanGotsiy: Could you move this array to https://github.com/Rebilly/ReDoc/blob/master/src/utils/openapi.ts

Also, having it as an object would be a bit faster than `array.includes` lookup.

src/components/Fields/Extensions.tsx
```diff
@@ -0,0 +1,46 @@
...
+    'x-nullable',
+    'x-servers',
+    'x-tagGroups',
+    'x-traitTag',
+  ];
+
+  render() {
+    const { schema } = this.props;
+    const fullSchema = schema.schema;
+    const extensionList = Object.keys(fullSchema).filter(
```
RomanGotsiy: Also, I think there is sense to do these filterings only when `options.showExtensions` is true.

I would add method called `getExtension`:

```ts
getExtensions() {
  const { schema } = this.props;
  const fullSchema = schema.schema;
  return Object.keys(fullSchema).filter(
    key => key.startsWith('x-') && !this.redocExtensions.includes(key),
  );
}
```

and then just used it in the `render` like bellow:

```ts
render() {
  return (
    <OptionsContext.Consumer>
      {options => (
        <>
          {options.showExtensions &&
            this.getExtensions().map(key => (
              <FieldDetail key={key} label={key} value={fullSchema[key]} />
            ))}
        </>
      )}
    </OptionsContext.Consumer>
  );
}
```

ingoclaro:
thanks for the feedback, I refactored the code.

ingoclaro:
@RomanGotsiy ping

RomanGotsiy:
@ingoclaro sorry for the delay!
Super busy...

Thanks for your efforts!

I will review it ASAP. Most likely at the beginning of the next week.

Please, ping me again if I disappear 🙈

ingoclaro:
resolved conflicts. ping @RomanGotsiy

RomanGotsiy:
I pushed a commit on top of this PR:
- implemented filtering by name (i.e. you can provide array of extensions names to show)
- added extensions to Operation
- extensions on `Parameter` object are now also included, not only on `Schema` Object
- changed look of extensions:
image

@ingoclaro Could you check it out and let me know what do you think?

ingoclaro:
@RomanGotsiy looks good to me. Thanks for the improvement.

jeremyfiel:
@RomanHotsiy @tatomyr    how can I use this in redocly cli with `preview-docs`?   no flag exists on this command for `--showExtensions`

tatomyr:
@AlexVarchuk  could you help with that?

AlexVarchuk:
Hi @jeremyfiel, you can create a config file, and pass it to the command
```
redocly preview-docs <api> [--config=<path>] [--port=<value>] [--host=<host>]
```

jeremyfiel:
What is the config attribute name?

PR #553, opened 2018-06-29
use original tag name when slugified one is not valid

mazicky:
The `slugify` function cannot return valid result when input is Unicode such as Korean, Chinese.. etc., so this occurs tag navigation and duplicated key error.


mazicky:
@RomanGotsiy I wrote wrapper function and use it at all usages. Please check it out.
Thanks.

RomanGotsiy:
Nice!
Thanks for your efforts 🙌!

Issue #550, opened 2018-06-28
externalDocs missing from tags, schemata and operations.


m-mohr:
I have this OpenAPI 3.0 specification:
```
...
    "/processes": {
      "get": {
        "summary": "List supported processes.",
        "description": "The request asks the back-end for available processes and returns detailed process descriptions, including parameters and return values. Processes are described using the Functio specification for language-agnostic process descriptions.",
        "tags": [
          "Process Discovery"
        ],
        "externalDocs": {
          "description": "Functio specification, version 0.1.0 (or any newer patch version).",
          "url": "https://github.com/m-mohr/functio-spec/blob/master/versions/0.1.0/spec.md"
        },
        "responses": {
          "200": {
            "description": "Functio document describing the supported processes.",
            "content": {
              "application/json": {
                "schema": {
                  "externalDocs": {
                    "description": "Functio specification, version 0.1.0 (or any newer patch version).",
                    "url": "https://github.com/m-mohr/functio-spec/blob/master/versions/0.1.0/spec.md"
                  }
                },
                "examples": {
                  "functioSpecExample": {
                    "externalValue": "https://raw.githubusercontent.com/m-mohr/functio-spec/master/versions/0.1.0/example.json"
                  }
                }
              }
            }
          },
          "4XX": {
            "$ref": "#/components/responses/client_error_auth"
          },
          "5XX": {
            "$ref": "#/components/responses/server_error"
          }
        }
      }
    },
...
```

Both externalDocs are not rendered by ReDoc as you can see here:
https://open-eo.github.io/openeo-api/v/0.3.0/apireference/index.html#/paths/~1processes/get

Would be helpful to have this feature implemented.

m-mohr:
Do you think this will be considered soon? Otherwise I'd need to start adding the externalDocs in the descriptions using markdown.

RomanGotsiy:
Yes, it will.
PR's are welcome!

m-mohr:
@RomanGotsiy I created a PR (#595) to solve this issue. May need some further tweaking regarding styling, but would needs some hints for that. Please feel free to review and add feedback as it was my first time working with React and there may be some issues.

matthunter:
I am also interested in this feature.  Is there timeline when this will get in?

RomanGotsiy:
I will review @m-mohr's PR ASAP. Hopefully this week.

m-mohr:
[...]

PR #595, opened 2018-08-06
External Documentation rendered for groups, operations and schemata.

m-mohr:
The externalDocs were missing for tags, schemata and operations. Added them with this pull requests. Solves #550.

Additionally, fixes that the URL in External Documentation Object was specified to be optional, which is not correct according to OpenAPI spec.


src/components/ContentItems/ContentItems.tsx
```diff
@@ -78,6 +79,15 @@ export class SectionItem extends React.Component<ContentItemProps> {
           </MiddlePanel>
         </Row>
         <AdvancedMarkdown source={description || ''} htmlWrap={middlePanelWrap} />
+        {externalDocs && (
+          <Row>
+            <MiddlePanel>
+              <p>
```
RomanGotsiy: Do we really need a `<p>` here?

src/components/Operation/Operation.tsx
```diff
@@ -50,6 +51,11 @@ export class Operation extends React.Component<OperationProps> {
               </H2>
               {options.pathInMiddlePanel && <Endpoint operation={operation} inverted={true} />}
               {description !== undefined && <Description source={description} />}
+              {externalDocs && (
+                <p>
```
RomanGotsiy: Same here

src/components/ContentItems/ContentItems.tsx
m-mohr: That's on purpose. The p adds a margin between the External Documentation link and the description. Without the margin in looks odd. The div is chosen in the schema as it seems to be designed to be very small and therefore without margin, like most other elements in there, too. So I think your suggestion makes things look worse without introducing additional styling per place that it's added.

src/components/Operation/Operation.tsx
m-mohr: See above.

RomanGotsiy:
I cleaned it up a bit myself!

Thanks for the PR.

It will land in the upcoming alpha at the end of the week.

m-mohr:
Thank you!

Issue #596, opened 2018-08-06
additionalProperties not rendered


m-mohr:
In schemata for objects it is allowed to set `additionalProperties` to `true` and allow arbitrary key-value pairs. Unfortunately this is not rendered by ReDoc and users reading the rendered documentation are not aware of the fact that they can add more properties than the one specified directly in the OpenAPI specification.

Example:
```
openapi: 3.0.1
info:
  title: Test
  version: "0.1.0"
paths:
  /example:
    get:
      summary: Example operation
      description: This is a simple example.
      responses:
        '200':
          description: This is a sample schema for a response with additionalProperties.
          content:
            application/json:
              schema:
                type: object
                properties:
                  id:
                    type: number
                  stringProperties:
                    type: object
                    additionalProperties:
                      type: string
                  untypedProperties:
                    type: object
                    additionalProperties: true
              example:
                id: 123
                stringProperties:
                  anotherProp: "test"
                untypedProperties:
                  anotherProp: "test"
```
Rendered as: https://rebilly.github.io/ReDoc/?url=http://mamo-net.de/feature/redoc/additionalprops.yaml

I would expect that additional properties of any type are allowed for the `untypedProperties` object. Should be the same way as for the `stringProperties`.

m-mohr:
Created a PR (597) to solve this issue.

PR #597, opened 2018-08-07
Bugfix that enables rendering additionalProperties set to true.

m-mohr:
Solves issue #596.

### DIFF
--- a/src/services/models/Schema.ts
+++ b/src/services/models/Schema.ts
@@ -244,7 +244,7 @@ function buildFields(
     sortByRequired(fields, schema.required);
   }
 
-  if (typeof additionalProps === 'object') {
+  if (typeof additionalProps === 'object' || typeof additionalProps === 'boolean') {
     fields.push(
       new FieldModel(
         parser,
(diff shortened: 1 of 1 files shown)


src/services/models/Schema.ts
```diff
@@ -244,7 +244,7 @@ function buildFields(
     sortByRequired(fields, schema.required);
   }
 
-  if (typeof additionalProps === 'object') {
+  if (typeof additionalProps === 'object' || typeof additionalProps === 'boolean') {
```
RomanGotsiy: if `additionalProperties === false` you should not push "additional properties" field, change this to checking if it is implicit `true`:

```js
if (typeof additionalProps === 'object' || additionalProps === true) {
```

m-mohr:
@RomanGotsiy Yes, that makes sense! Thanks for the review. Applied your suggestions, please re-check.

src/services/models/Schema.ts
m-mohr: Implemented.

RomanGotsiy:
Nice 👍
Thanks!

PR #614, opened 2018-08-22
Feature/dockerize redoc cli

davidoram:
This change packages `redoc-cli` as a docker image. Its useful to have the `redoc-cli` tool pre-packaged with all its dependencies if you want to use it in your toolchain.  

To build the image:

```
$ cd <Redoc project directory>
$ docker build -t redoc-cli -f cli/Dockerfile .

```

and then to run it :

```
$ docker run --rm -it  -v $PWD:/data redoc-cli redoc-cli bundle swagger.yml
```


cli/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN apk update && apk add --no-cache git python alpine-sdk nodejs
+
+WORKDIR /build
+COPY . /build
+
+RUN yarn install --frozen-lockfile --ignore-optional --ignore-scripts && \
+  npm run bundle && npm run compile:cli && \
+  npm install typescript && \
+  npm install iltorb  && \
+  npm install -g redoc-cli
```
RomanGotsiy: I don't understand. You first build it from source and then install from npm. What's the point?

cli/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# $ docker run --rm -it redoc-cli redoc-cli --help
+# .. will display the comand line help
+#
+# To the `swagger.yml` file in the current directory, and output 'redoc-static.html'
+# $  docker run --rm -it  -v $PWD:/data redoc-cli redoc-cli bundle swagger.yml
+
+FROM node:alpine
+
+# Needs python, alpine-sdk to build node packages
+RUN apk update && apk add --no-cache git python alpine-sdk nodejs
```
xxxcoltxxx: To minimize image size all temporary files must be deleted in the same layer (in current `RUN` command

cli/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY . /build
+
+RUN yarn install --frozen-lockfile --ignore-optional --ignore-scripts && \
+  npm run bundle && npm run compile:cli && \
+  npm install typescript && \
+  npm install iltorb  && \
+  npm install -g redoc-cli
+
+WORKDIR /data
+
```
xxxcoltxxx: PLease add an `ENTRYPOINT` and `EXPOSE` commands

davidoram:
@RomanGotsiy Good question.  In my use case I want to use the `redoc-cli` tool in my toolchain, but using a stable, repeatable process.

In my projects `Makefile` I build my html docs by running `redoc-cli`  inside a docker container, using it a bit like a command line tool eg:

`docker run --rm -it  -v $(CURDIR):/data davidoram/redoc-cli redoc-cli bundle swagger.yml ./path/to/output/`

This will generate my html docs with no special installation apart from docker itself.

cli/Dockerfile
RomanGotsiy: I see.
Then I think you overcomplicated. All you need is a Dockerfile as simple as bellow:

```
FROM node:alpine

RUN npm install -g redoc-cli
```

Could you check if it works?

andrdru:
@RomanGotsiy  @davidoram 

I think dockerfile should be like this

```
FROM node:alpine

RUN apk update && apk add --no-cache nodejs &&\
    npm install -g redoc-cli

WORKDIR /data
EXPOSE 8080
ENTRYPOINT ["redoc-cli"]
CMD []
```

```
docker build -t redoc .
```

generating bundle:
```
docker run --rm -v $PWD:/data redoc bundle swagger.yml
```

serving:
```
docker run --rm -v $PWD:/data -p 8080:8080 redoc serve swagger.yml
```

davidoram:
@RomanGotsiy I tried your suggestion and couldn't get that to work unfortunately. When I run that I get the following:

```
docker run --rm -v $PWD:/data b4c14bfbaaad bundle swagger.yml
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"bundle\": executable file not found in $PATH": unknown.
```

I'm not a docker expert. Is docker interpreting the command line option as the command itself?

davidoram:
@RomanGotsiy I _think_ that I have applied all the changes reqeested, or am I missing something?

RomanGotsiy:
@davidoram looks good! Thanks!

PR #619, opened 2018-08-23
Update readme color customization example

kmbenitez:
Setting options.theme.colors.main does not work. Digging in the options object revealed that it is options.theme.colors.primary.main.


RomanGotsiy:
Thanks, @kmbenitez!

PR #620, opened 2018-08-23
update cli dependencies

ramitos:


cli/package.json
```diff
@@ -11,12 +11,12 @@
     "handlebars": "^4.0.11",
     "isarray": "^2.0.4",
     "mkdirp": "^0.5.1",
-    "mobx": "^4.2.0",
-    "react": "^16.3.2",
-    "react-dom": "^16.3.2",
-    "redoc": "^2.0.0-alpha.29",
-    "tslib": "^1.9.0",
-    "yargs": "^11.0.0"
+    "mobx": "^5.0.4",
```
RomanGotsiy: We can't update to mobx@5 because it is based on js Proxies which can't be polyfilled and have not so good browsers support.

Please undo this change

ramitos:
@RomanGotsiy 👍

PR #632, opened 2018-09-03
Increase badge size slightly so that "PATCH" method fits inside better

fredriklengstrand:
This pull request fixes an issue where the "PATCH" method does not fit inside the badge very nicely at the cost of some horizontal space for the summary.

**Before:**

screen shot 2018-09-03 at 20 45 23

**After:**

screen shot 2018-09-03 at 20 40 03


RomanGotsiy:
Hi @fredriklengstrand, thanks for the PR.

I don't think increasing badge size is a good choice. This badge takes too much space even now.

I would prefer shortening `PATCH` to something like `PTCH` (similar to how it is done for `DELETE` -> `DEL`) or making `PATCH` font size a bit smaller.

What do you think?

RomanGotsiy:
I tried both shortening `PATCH` to something like PTCH and making `PATCH` font size smaller and both look not so good.

Looks like making badge size bigger is the best option here.

PR #663, opened 2018-10-03
Response sample languages

manuquentin:
I appreciate the `x-code-samples` feature of ReDoc. But in my case I've a Javascript SDK that returns something different than my API (because the SDK returns Javascript class instances).

So I would like to add a `x-response-samples` that will be added to the list of http status like this :

response-code

In that case we'll be able to return code in addition to API responses.


RomanGotsiy:
This is a bit inconsistent.
You put `200` which is a response code and `JavaScript` into one row.

I understand your use case but I can't merge it as is. I also don't have any ideas on how to make it consistent.

Do you have some ideas?

RomanGotsiy:
@manuquentin ping

RomanGotsiy:
Sorry but have to close it as the solution proposed is inconsistent.

Feel free to suggest some alternative solution.

PR #675, opened 2018-10-18
service/AppStore: add null check in dispose method

williamboman:
The `disposer` field were of `any` type so TS did not catch this.

If `disableSearch` option is set to true, we never set up the mobx observer, effectively leading to `this.disposer` being `undefined`.

### DIFF
--- a/src/services/AppStore.ts
+++ b/src/services/AppStore.ts
@@ -1,4 +1,4 @@
-import { observe } from 'mobx';
+import { observe, Lambda } from 'mobx';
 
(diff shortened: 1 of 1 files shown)


RomanGotsiy:
@williamboman good catch 👍
Thanks!

williamboman:
No probs! Would it be possible to get a new release with this included 🙈?

RomanGotsiy:
It's in progress: https://travis-ci.org/Rebilly/ReDoc/builds/443139170

PR #691, opened 2018-10-27
Corrected Docker Port in README.md

RobertDeRose:
The command to launch ReDoc with Docker had the `-p` values reversed.
According to the Docker Docs, it should be `HOST_PORT:CLIENT_PORT`.
The Nginx service is running on port 80 inside the container, 8080 is the 
port exposed on the host.


RomanGotsiy:
Thanks!

PR #697, opened 2018-11-03
refactor(cli): Use chokidar for watching

mes3yd:
Caveat: could not get this to work with a debounce.
Fixes: recursive file watching on Linux (and thus Docker).


RomanGotsiy:
@mes3yd LGTM 👍 
Could you please just resolve cli/yarn.lock conflicts? I think you can just remove and run `yarn install` to generate a clean `yarn.lock` file.

PR #708, opened 2018-11-11
Fix #694: Add READ MORE for long description

### LINKED ISSUE #694: Suggestion: "Read more" button for long description
Sometimes `description` field of API and/or Component can be quite long.
I think long description makes browsing API Doc browsy a little.

It would be helpful if we can split `description` into "always-shown section" and "more section", and add `Read more` button to toggle "more section" visibility.

Something like this.

```yaml
/foo/{bar}:
  get:
    summary: get Bar from Foo
    description: |
       Yay.

       These lines are visible always.
       <!-- READMORE -->
       These lines are visible only if "Read More" is enabled.
       ...very very very long
```

exoego:
# Usage

Add a special comment `<!-- READMORE -->` in `description`.
More content fade in from white, so users can notice there is more content.

```
/foo/{bar}:
  get:
    summary: get Bar from Foo
    description: |
       These lines are visible always.
       <!-- READMORE -->
       These lines are visible only if "Read More" wrapper is open.
       ...very very very long
       <!-- READMORE --> ⬅️ Second and later occurences are treated as normal comment.
```

# Screenshot
read-more

# Closes
Closes https://github.com/Rebilly/ReDoc/issues/694

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,9 @@ Also you may rewrite some predefined environment variables defined in [Dockerfil
 ### Security Definition location
 You can inject Security Definitions widget into any place of your specification `description`. Check out details [here](docs/security-definitions-injection.md).
 
+### "Read More" button
+You can add "Read More" button that allows readers to continue reading more by adding `<!-- READMORE -->` into any place of your specification `description`. Check out details [here](docs/read-more-description.md).
+
 ### Swagger vendor extensions
 ReDoc makes use of the following [vendor extensions](http://swagger.io/specification/#vendorExtensions):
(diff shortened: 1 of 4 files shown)


RomanGotsiy:
@exoego sorry for long delay.
It looks awesome! Thanks!

Although I am not sure about the "read more" button styling.
I am quite busy now so will get to this later.

exoego:
Closing since browser's native support of details/summary tag.
https://github.com/Redocly/redoc/issues/694#issuecomment-531659295

PR #716, opened 2018-11-16
refactor(cli): Update watching:

mes3yd:
- For now hardcode ignore JetBrain's "safe file changes".
- Show events for new files and directories
- Update spec on new files

Considering:
- Put the ignore in package.json as a config value or a separate
  config file that fills watchOptions.


cli/index.ts
```diff
@@ -170,7 +170,8 @@ async function serve(port: number, pathToSpec: string, options: Options = {}) {
   if (options.watch && existsSync(pathToSpec)) {
     const pathToSpecDirectory = resolve(dirname(pathToSpec));
     const watchOptions = {
-      ignored: /(^|[\/\\])\../,
+      ignored: [/(^|[\/\\])\../, /___jb_[a-z]+___$/],
```
RomanGotsiy: Could you add a comment explaining the second regexp?

cli/index.ts
```diff
@@ -184,7 +185,21 @@ async function serve(port: number, pathToSpec: string, options: Options = {}) {
           log('Updated successfully');
         } catch (e) {
           console.error('Error while updating: ', e.message);
-        }})
+        }
+      })
+      .on('add', async path => {
+        log(`File ${path} added, updating docs`);
+        try {
```
RomanGotsiy: These five lines are exactly the same as above in `change` handler. I think there is a point to create a function

RomanGotsiy:
Hey @mes3yd. Would you have a chance to update this PR according to comments?
It would be awesome!

Thanks in advance!

RomanHotsiy:
pink @mes3yd

PR #725, opened 2018-11-27
chore: use tarball url for marked dependency

lpinca:
Allow the dependency to be installed without git.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "json-schema-ref-parser": "^6.0.1",
     "lunr": "^2.3.2",
     "mark.js": "^8.11.1",
-    "marked": "https://github.com/markedjs/marked#fb48827",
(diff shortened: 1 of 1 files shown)


lpinca:
I wonder if it's possible to upgrade to marked@0.5.2 instead, it would be better.

RomanGotsiy:
Sorry for the long reply
see https://github.com/Rebilly/ReDoc/pull/734

lpinca:
Ok, can we use a tarball url in the meanwhile with your latest fix to marked?

lpinca:
I mean do they have an ETA for the next release? If not it might makes sense.

lpinca:
It would be great. It's not strictly a blocker but some of my coworker are relying on horrible hacks due to this git dep :)

PR #728, opened 2018-11-29
Fix links to OAS spec page

patrickelam:
It looks like the OAS spec page changed some of their anchor tags, just fixing the links that I noticed.


RomanGotsiy:
Thanks! 👍

PR #768, opened 2019-01-02
chore: update marked to version 0.6.0

lpinca:
The new release includes https://github.com/markedjs/marked/pull/1383.
Would it be possible to also release a new version of redoc and redoc-cli?

Thank you.


lpinca:
Opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31834 to address the failures, will update accordingly here when possible.

ChrisDillinger:
Hi there! Can this PR be merged? What else needs to be reviewed so that this PR can be successfully closed out? I am very much looking forward to this change here: https://github.com/Rebilly/ReDoc/commit/809e7f2841b4cb034e8d23e92a3ebfef23580022

As I really want to integrate this wonderful open source tool into my project. Thanks so much guys for being such a cool community.

src/services/MarkdownRenderer.ts
```diff
@@ -129,7 +131,7 @@ export class MarkdownRenderer {
         this.currentTopHeading && this.currentTopHeading.id,
       );
     }
-    return this.originalHeadingRule(text, level, raw);
+    return this.originalHeadingRule(text, level, raw, this.slugger);
```
RomanGotsiy: Looks like you don't need to manually create `slugger`.
Just pass it through function parameters:

```ts
headingRule = (text: string, level: number, raw: string, slugger: marked.Slugger) => {
  // ...
  return this.originalHeadingRule(text, level, raw, slugger);
}
```

lpinca: Good catch, addressed.

RomanGotsiy:
Thanks!

PR #773, opened 2019-01-08
Only display API version if present

TheMarex:
This only displays the version if it is specified in the info section:

## Before 

image

## After 

image

This is useful for OpenAPI specifications that bundle several APIs that are versioned independently, hence the global version string makes little sense.

### DIFF
--- a/src/components/ApiInfo/ApiInfo.tsx
+++ b/src/components/ApiInfo/ApiInfo.tsx
@@ -70,12 +70,18 @@ export class ApiInfo extends React.Component<ApiInfoProps> {
       )) ||
       null;
 
+    const version =
+      (info.version && (
+        <span>({info.version})</span>
+      )) ||
+      null;
+
     return (
       <Section>
         <Row>
           <MiddlePanel className="api-info">
             <ApiHeader>
-              {info.title} <span>({info.version})</span>
+              {info.title} {version}
             </ApiHeader>
             {!hideDownloadButton && (
               <p>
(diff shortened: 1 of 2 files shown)


RomanGotsiy:
`version` field is required according to [the OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-1):

image

TheMarex:
Agreed, the implementation should conform to specs. However, it seems to me that ReDoc generally prefers to degrade in service, rather than fail if it encounters a non-compliant spec.
Given how prevalent per-operation versions are, it is questionable if the OpenAPI standard should have made this a required field in the first place, but this is a discussion for another day.

This changes simply ensures that this degradation of service is more resilient. Given how simple this change this, and that it fixes a real issue, I would advocate for merging this. If your concern is that it makes a non-compliant spec less easy to catch, I can add a warning so it remains discoverable to developers.

RomanGotsiy:
> However, it seems to me that ReDoc generally prefers to degrade in service, rather than fail if it encounters a non-compliant spec.

@TheMarex agree! Just add a warning, please.

Thanks!

TheMarex:
@RomanGotsiy done!

RomanGotsiy:
nice! Thanks!

RomanGotsiy:
In the end, decided to remove the warning. ReDoc is documentation tool not validator :)

PR #776, opened 2019-01-09
Incorrect port after restart when PORT is 8080

DeBr0glie:
When PORT is 8080 and dokcer restart, nginx will listen 80808080


RomanGotsiy:
Thanks!

PR #778, opened 2019-01-09
Fix the demo specUrl

stackedsax:
Rebilly's `swagger.json `was renamed to `openapi.json`.  Just checking in a quick fix for the change.


RomanGotsiy:
Thanks!

PR #805, opened 2019-02-04
Fix typings, because `compile:cli` fails with errors

bautrukevich:


RomanGotsiy:
Fixed it with TypeScript without runtime conversions in https://github.com/Rebilly/ReDoc/commit/f31cf3534c09b5e7f0e003a2b5b4328cd65daab2

Thanks for the PR anyway!

Issue #822, opened 2019-02-26
Escaped double-quote \" is not shown as escaped in example


exoego:
* Tested on https://cdn.jsdelivr.net/npm/redoc@2.0.0-rc.2/bundles/redoc.standalone.js

My schema has a string field which stores JSON template like

```yml
body:
    type: string
    example: "{\"message\": \"test\"}"
```

or

```yml
body:
    type: string
    example: '{"message": "test"}'
```

But both of them are shown like

![image](https://user-images.githubusercontent.com/127635/53389360-55fbee00-39d2-11e9-83fe-1d6549acf0ef.png)

where double-quote `"` is shown directly, therefore the response exmple looks like broken.

I think doubel-quote are escaped and shown like `\"`.

exoego:
Closing as self-resolved.

We should write example with escaping like this.

```
example: '{\\"message\\": \\"test\\"}'
```

![image](https://user-images.githubusercontent.com/127635/53389695-a32c8f80-39d3-11e9-87ae-41af32ec6980.png)

exoego:
I think this is issue. Escape should be with single `\`, not `\\'

```
    example: "{\"message\": \"test\"}"
```

exoego:
Sorry many times.

```json
` example: "{\"message\": \"test\"}"`
```

ReDoc can handle the above.

![image](https://user-images.githubusercontent.com/127635/53397247-01fd0380-39ea-11e9-8219-84466f9971ed.png)

PR #823, opened 2019-02-26
fix: escape backslashes in string literals

unarist:
Follow up for #822 and 04731656 (only escape quotes).

When `example: '{"name":"Guru \ bs"}'` is given...

before:
image

after:
image


RomanGotsiy:
Nice! Thanks!

PR #841, opened 2019-03-11
Fixed docker README: added missing redocly repo name

julmuell:
By just using the commands from https://hub.docker.com/r/redocly/redoc/ it didn't worked for me. I added the missing docker repository name and changed PWD to pwd as well.


RomanGotsiy:
Thanks!

Issue #707, opened 2018-11-10
Support request samples for JSON parts in multipart/form-data requests


gyszalai:
We have an API with an endpoint that handles file uploads. The endpoint accepts multipart/form-data requests with two parts being a json content in the first part and binary file data in the second part.
We would like to give a request sample to the first (JSON) part. 

We use openapi 3.0.0 spec, the request definition looks like this:

```yaml
requestBody:
    required: true
    content:
        multipart/form-data:
            schema:
                type: object
                properties:
                    meta:
                        $ref: './meta.schema.json'
                    attachment:
                        type: string
                        format: binary
                        required: true
                required:
                - meta
                - attachment
            encoding:
                meta:
                    contentType: application/json; charset=utf-8
```
The meta schema contains an example, and it renders correctly if the request body is application/json.
Are we doing something wrong or ReDoc does not support our use case? Thanks!

RomanGotsiy:
You're doing nothing wrong. ReDoc just doesn't support your use case yet.

bautrukevich:
@RomanGotsiy Can I make PR to fix this issue? Because I need this feature too. Please, can you explain, what is the problem with it? And where I can fix that?
Thank you.

jthiel1996:
I am seeing the same issue described by gyszalai on Nov 10, 2018
Here is what i tried to do:
1.  npm install redoc --save . To reinstall Redoc.  I still get the error.
2.  Attempted to use code from CDN (<script src="https://cdn.jsdelivr.net/npm/redoc/bundles/redoc.standalone.js"> </script>) assuming that this had latest.

Is there a third way that I need to do this to get the fix?

jessevdp:
I noticed that the PR that would supposedly implement support for this was closed without merging. Any reason for that?

Simon-Gregory-LG:
I'll add my voice to those that think multipart support is a must

Claudezss:
A quick solution to show example of multipart/form-data:
on `src/components/PayloadSamples/MediaTypeSamples.tsx` line 41, change `if (examplesNames.length > 1)` to `if (examplesNames.length > 0)`

sample result:

<img width="539" alt="Screen Shot 2021-04-10 at 5 52 23 PM" src="https://user-images.githubusercontent.com/33612092/114285519-c08f0d00-9a25-11eb-81b9-63fd831da552.png">

agaertner:
We are using ReDoc and would like to see this supported as well. Multipart/form-data is very common.

[...]

PR #846, opened 2019-03-14
Fix error with multipart/form-data

bautrukevich:
Related to https://github.com/Rebilly/ReDoc/issues/707

### DIFF
--- a/src/components/PayloadSamples/Example.tsx
+++ b/src/components/PayloadSamples/Example.tsx
@@ -2,7 +2,8 @@ import * as React from 'react';
 
 import { StyledPre } from '../../common-elements/samples';
 import { ExampleModel } from '../../services/models';
-import { isJsonLike, langFromMime } from '../../utils';
+import { isJsonLike, isTextPlainLike, langFromMime } from '../../utils';
+import { jsonToTextPlain } from '../../utils/jsonToTextPlain';
 import { JsonViewer } from '../JsonViewer/JsonViewer';
 import { SourceCodeWithCopy } from '../SourceCode/SourceCode';
 import { ExampleValue } from './ExampleValue';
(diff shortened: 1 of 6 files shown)


RomanGotsiy:
Thanks for your PR!
I looked through it quickly and it looks good in general.

Although, I have a few concerns I want to investigate before merging this in.
If I disappear 🙈  keep pinging me, please 👉.

bautrukevich:
@RomanGotsiy maybe you want to leave some comments? I will be glad. Because I need this feature as soon as possible. And can we release a patch with this?

RomanGotsiy:
OK. So the first question is: why do you encode multipart/form-data as key-values pairs?

It seems to not reflect how things are encoded in reality so this is confusing.

Also, you say it is related to #707 but I look like this PR doesn't fix the issue.

> Because I need this feature as soon as possible.

If it doesn't land soon enough you can just use your fork temporary.

bautrukevich:
@RomanGotsiy yes, you are right, this is not how it looks in browsers requests, or not looks like HTTP request, and it's simplified because we need to show somehow what developer need to send in the request.
About issue — I am using my own example and it works fine if I am setting requestBody content as multipart/form-data.
About fork — how can I use them in redoc-cli? Because redoc-cli has own package.json and do I need to use fork of redoc-cli and after one more fork for redoc? 
Thank you!

g0ddish:
Could we get this rolled up into a release by any chance 🙈

ClaudiaSaft:
Any update on this? or a recommended workaround/alternative ?

@bautrukevich Why did you delete your fix branch? Any specific reason?

Thanks!

bautrukevich:
@ClaudiaSaft I don't remember exactly, maybe because it not actual anymore

PR #847, opened 2019-03-14
Tidy up non-redoc extension presentation

drjonnicholson:
This PR:

- Standardises an extension to use the `FieldLabel` component rather than a bespoke one so the style matches
- Removes the `x-` prefix from the extension when presenting the name on the page
- Title cases the unprefixed extension name
- Removes the extraneous `"` quotes from either side of the extension value, so strings like `test` are displayed as `test` and not `"test"`

Before:
image
(where we're using `x-scope` to annotate the scope required for a particular field to be populated from our API above and beyond the scope needed to access the endpoint itself, which is defined in security objects etc)

After:
image

### DIFF
--- a/src/components/Fields/Extensions.tsx
+++ b/src/components/Fields/Extensions.tsx
@@ -1,4 +1,9 @@
 import * as React from 'react';
+
+import { startCase } from 'lodash';
+
+import { FieldLabel } from '../../common-elements/fields';
+
 import styled from '../../styled-components';
 
 import { OptionsContext } from '../OptionsProvider';
@@ -8,10 +13,6 @@ import { StyledMarkdownBlock } from '../Markdown/styled.elements';
 const Extension = styled(StyledMarkdownBlock)`
   opacity: 0.9;
   margin: 2px 0;
-`;
-
-const ExtensionLable = styled.span`
-  font-style: italic;
 `;
 
 export interface ExtensionsProps {
@@ -29,8 +30,10 @@ export class Extensions extends React.PureComponent<ExtensionsProps> {
             {options.showExtensions &&
               Object.keys(this.props.extensions).map(key => (
                 <Extension key={key}>
-                  <ExtensionLable>{key}</ExtensionLable>:{' '}
-                  <code>{JSON.stringify(this.props.extensions[key])}</code>
+                  <FieldLabel>{startCase(key.substring(2))}</FieldLabel>:{' '}
+                  <code>
+                    {JSON.stringify(this.props.extensions[key]).replace(/(^")|("$)/g, '')}
+                  </code>
                 </Extension>
               ))}
           </>
--- a/src/components/Fields/Extensions.tsx
+++ b/src/components/Fields/Extensions.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { FieldLabel } from '../../common-elements/fields';
+import { FieldLabel, ExtensionValue } from '../../common-elements/fields';
 
 import styled from '../../styled-components';
 
@@ -9,7 +9,6 @@ import { OptionsContext } from '../OptionsProvider';
 import { StyledMarkdownBlock } from '../Markdown/styled.elements';
 
 const Extension = styled(StyledMarkdownBlock)`
-  opacity: 0.9;
   margin: 2px 0;
 `;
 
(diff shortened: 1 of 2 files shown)


drjonnicholson:
As an aside, I ran `npm run prettier` before making the commit to make sure it met code standards, and it changed a bunch of files

src/components/Fields/Extensions.tsx
```diff
@@ -1,4 +1,9 @@
 import * as React from 'react';
+
+import { startCase } from 'lodash';
```
RomanGotsiy: Is there a point to pull in lodash because of the only `startCase` function?

Could you implement a simple one yourself e.g. in `utils` or just copy lodash's one

RomanGotsiy:
Also, add a screenshot of how it looks like now, please! That way it would be much easier for me.

Thanks in advance!

src/components/Fields/Extensions.tsx
exoego: Why are you afraid of importing single function ?
I don't think it is good idea to re-invent (or copy) common functions already defined in libralies.

RomanGotsiy: It's not about importing a single function. It's about adding a new dependency as `lodash` is not in the dependencies yet.

The main reasoning behind this is the bundle size. While it doesn't matter so much for node apps it matters for browsers. Tree-shaking isn't perfect. I just tested and importing this single function leads to *68KB!* increase in bundle size. Don't you find it too much for such a simple function (especially since vendor extension names can't contain any special Unicode symbols).

Hope this does make sense.

exoego: It makes sense 👍 
Looks like `lodash` are not used in production code right now, so importing single function may add `68KB' or around to bundle.
image

Perhaps `lodash` dependencies should be replaced with more granular modules like [lodash.set](https://www.npmjs.com/package/lodash.set).

RomanGotsiy: Yes, but this is development code so it doesn't matter so much and I don't think it's worth to do.

drjonnicholson: Hi @RomanGotsiy and @exoego, thanks for having a look at this.

So I agree with @exoego to avoid reinventing the wheel etc where possible, but 68k bundle increase is too big for a quite simple function. I'm ultimately happy with switching this out to something more appropriate. Happy with what you guys decide really.

drjonnicholson:
@RomanGotsiy I've updated the PR description with before and after images

src/components/Fields/Extensions.tsx
RomanGotsiy: I'm not sure now if we need to title-case the extension name at all.

Vendor extension can contain arbitrary name so e.g.: `x-Some Title` is a valid extension. So you have a workaround to show title-cased labels but if we title case them by default other users don't have a way to display lower-case label.

Summing up, I think we should remove `startCase` at all. Do you agree?

drjonnicholson: Fair point, if `x-Some Title` is valid then yeah completely agree - we can't guess or impose one method of presentation over another.

And I can solve my use case by just capitalising the first letter after the `x-` prefix, so I'm happy.

I'll update the PR in a moment

RomanGotsiy:
Thanks! While you're already on it. Could you plz also check text vertical alignment issue (check on the screenshot) 🙇 

If you can't fix it in 10-15 mins, left it for me!

drjonnicholson:
Good spot, so looking at default/exmaples, the `:` is inside the `FieldLabel` component, so that solves the vertical alignment issue.

There is also a horizontal issue in the value being closer to the label than it is to say an example. Here's a mocked up example:
image

I could either:
- Leave as is
- Reuse the existing `ExampleValue` styled component
- Duplicate the `ExampleValue` component as `ExtensionValue` in `field.ts`, giving us a new extension hook etc.

Personally I like the duplication in this case - since it's initially duplicated but can be modified independently of examples etc, which serve a different purpose

I'm also tempted to remove the opacity of the `Extension` component. That sound good?

Any other changes while I'm at it? 😃

RomanGotsiy:
> Personally I like the duplication in this case - since it's initially duplicated but can be modified independently of examples etc, which serve a different purpose

Can you try extending just extending it? i.e. 

```js
export const ExtensionValue = styled(ExampleValue)``
```

No extra extension hook though but it can be added later if needed.

> I'm also tempted to remove the opacity of the Extension component. That sound good?

Ye, go ahead!

drjonnicholson:
Yeah that makes sense. Next commit going up, let me know if there's anything else under this PR you want done.

drjonnicholson:
Think that's everything. Output of `Example` against and `Extension` now looks like mock up (just tweaking he DOM a little to put them next to each other with same label value of ` Example: `):

image

RomanGotsiy:
Looks awesome now! Thanks! 🙇

RomanGotsiy:
Just refactored it a bit https://github.com/Rebilly/ReDoc/commit/dc77d8f25e8d9f198eb1135b01eb5b04dc9555ec.

Check it out please just in case.

drjonnicholson:
Nice change. I imagine Inspecting the type must be faster than string manipulation. Feel bad I didn’t think of that approach!

Thanks for making a great product all, and for making submitting improvements a welcoming experience 🙂

Issue #826, opened 2019-02-27
[Feature Request] Is it possible to make x-tagGroups collapsible


lordrhodos:
I was wondering if there is any option to enable the tag groups to be collapsible and be collapse by default? 

The [documentation](https://github.com/Rebilly/ReDoc/blob/master/docs/redoc-vendor-extensions.md#x-tagGroups) does not show any hints about this and I can not find anything in the repository about it. 

We have very large collection of endpoints and it would improve the usability a lot.

If this is not supported yet I am happy to look into this and provide a PR to add this feature.

RomanGotsiy:
This is not supported. Your PR would be appreciated!

bradennapier:
@lordrhodos i put this together day when i needed it and submitted a PR.  Let me know if it isn't what you were thinking!

lordrhodos:
@bradennapier thanks for your effort on this issue. 

I started a couple of days ago on an own implementation and I chose to follow a slightly different approach. I introduced a new configuration parameter `collapseTagGroups` which defaults to `false` leaving everything as is now. 

If set to `true` it would add collapseble and expandable feature on the model and use a different styling. I still need to figure out some injection parts in the `styled.elements.ts` components though. I am not a React/Typescript pro ;-)

I like your approach adding the collapsible support for groups on a single group level, although I prefer to have only one central configuration switch. And I like the visual integration of the collabsible group menu item, but I am not sure if we need a highlight for active (expanded) groups. This would also need some adjustments for the padding I guess as changing the background color for highlighting does not look nice.

Will look into this soon and be back with some more feedback.

bradennapier:
Can you expand on what you mean with the highlighting issue? Not sure I see any issue there but it’s likely a very simple css thing.  

A whole new switch and model seems excessive.  This solutions primary benefit imo is that it actually needs to do very little to achieve the goal and doesn’t need to add any new concepts or configuration.

lordrhodos:
@bradennapier yes I agree this to rather a css issue. If you add highlighting via background colour for `:hover` and `active` state you probably want to change the bottom padding as well I guess.

> A whole new switch and model seems excessive. This solutions primary benefit imo is that it actually needs to do very little to achieve the goal and doesn’t need to add any new concepts or configuration.

I agree that a global switch is debatable. I rather prefer to set a configuration setting than duplicating a setting in the spec for `n` group tags ;-)

[...]

PR #853, opened 2019-03-19
[FEATURE] Collapsible Groups

bradennapier:
So I needed this so I figured I'd try my hand at contributing to the repo!  I put some time into thinking out the various potential cases to try to make the UX intuitive here.

1. `x-tagGroups` can have a setting `collapsible: true` which will make them a collapsible group.
2. Retains general scroll-through behavior (auto-expand as user scrolls, etc).
3. Allows closing / collapsing of a group by clicking on an open group.  
  - This will scroll to the first element of the group, scrolling down will then re-open the group naturally.
4. Collapsible groups default to `collapsed` unless you are viewing a subitem of the group or click the item.
5. Adds a property `collapsible` to tag items which defaults to `true`.  Uses this value along with `items.length` to determine if it is collapsible.
6. Adds margin to each level rendered (clearly collapsible -> tag -> operations has reduced width available).
7.  I would note that running prettier seems to have adjusted some other files that I did not touch in this PR.  It would seem it wasn't run on a previous PR or commit.  Just an FYI there.  
  - For easier diffing, [omitting whitespace may be best](https://github.com/Rebilly/ReDoc/pull/853/files?w=1)

---

### Resolves:

#826 

---

### Spec Example

```yaml
tags:
  - name: pet
    description: Everything about your Pets
  - name: store
    description: Access to Petstore orders
  - name: user
    description: Operations about user
  - name: SubItem One
    description: |
      SubItem One Description!
  - name: SubItem Two
    description: |
      SubItem Two Description!
x-tagGroups:
  - name: General
    collapsible: true
    tags:
      - pet
      - store
  - name: User Management
    tags:
      - user
  - name: Collapsible Group
    collapsible: true
    tags:
      - SubItem One
      - SubItem Two
```
---

### To Do:

- [x] Currently collapsing a 2-level collapsible will navigate to the element above it if collapsed and it isn't a top-level item.  Need to fix this.  

> General > [pet, store] -- When store is expanded and you collapse it at the moment it will scroll to the top of pet rather than the top of store.  Will fix this shortly.

- [ ] Was not sure if the `menu.collapse()` needed to be decorated with `@action` as I am not familiar with `mobx` and it does work fine without it.  If it is needed, prob easiest to just use the github suggestions feature and add it inline here so we can commit quickly once approved :-) 

- [x] **NOT PLANNED** - Technically operations do not add the proper margin to them so the method tag is rendered a bit too far to the left.  I did this on purpose to preserve space since it does not seem to look that bad.  Easy enough to provide if desired though.

- [x] **NOT IN SCOPE** - related to the above, for truly recursive handling of tags in this manner, we would need to do a more powerful search to find the best element to scroll to in this case.  It does not seem necessary as it is unlikely that supporting more than 2-levels will make sense at any point (or even be possible).  Honestly only real use-case I see for this is for better organizing of what would otherwise be top-level tags.

---

### Side Notes

- I noticed that if you happen to render the same tag in multiple groups it will break.  While in many cases this probably will never happen - a few things should probably be considered here (out of scope and not planning to submit PR for this one, just noting it while I am typing :-D):
  1. Groups should be rendered as top-level heading encapsulating its child items (perhaps optional via setting).  
      - This would also simplify scroll behavior since we do not need to try to determine best place to scroll to when a group is actually clicked.
  2. History should include the general breadcrumbs if possible for differentiating values.

---

### Screenshots

image

image

image

Also supports mixed collapsible/static:

image

### DIFF
--- a/demo/openapi.yaml
+++ b/demo/openapi.yaml
@@ -74,14 +74,14 @@ x-tagGroups:
     tags:
       - pet
       - store
+  - name: User Management
+    tags:
+      - user
   - name: Collapsible Group
     collapsible: true
     tags:
       - SubItem One
       - SubItem Two
-  # - name: User Management
-  #   tags:
-  #     - user
 
 paths:
   /pet:
--- a/demo/openapi.yaml
+++ b/demo/openapi.yaml
@@ -17,14 +17,14 @@ info:
     It was **extended** to illustrate features of [generator-openapi-repo](https://github.com/Rebilly/generator-openapi-repo)
     tool and [ReDoc](https://github.com/Rebilly/ReDoc) documentation. In addition to standard
     OpenAPI syntax we use a few [vendor extensions](https://github.com/Rebilly/ReDoc/blob/master/docs/redoc-vendor-extensions.md).
-    
+
     # OpenAPI Specification
     This API is documented in **OpenAPI format** and is based on
     [Petstore sample](http://petstore.swagger.io/) provided by [swagger.io](http://swagger.io) team.
     It was **extended** to illustrate features of [generator-openapi-repo](https://github.com/Rebilly/generator-openapi-repo)
     tool and [ReDoc](https://github.com/Rebilly/ReDoc) documentation. In addition to standard
     OpenAPI syntax we use a few [vendor extensions](https://github.com/Rebilly/ReDoc/blob/master/docs/redoc-vendor-extensions.md).
-    
+
     # Cross-Origin Resource Sharing
     This API features Cross-Origin Resource Sharing (CORS) implemented in compliance with  [W3C spec](https://www.w3.org/TR/cors/).
     And that allows cross-domain communication from the browser.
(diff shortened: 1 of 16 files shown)


bradennapier:
Tests are failing but it does not appear to be due to my changes here as others are seeing same.  I can't imagine this would cause tests to fail but possible ofc!

bradennapier:
Just shipped our documentation over at IDEX with this PR, https://docs.idex.market/  for reference

mikouaj:
+1
feature is useful when creating API catalog for bigger platforms

ngthtung:
+1

CarltonHowell:
@bradennapier Any chance this gets merged in production? Would love this feature!

bradennapier:
The team never even responded to it so I don't think they are really open to ideas or public contribution.  

At this point this PR is likely so out of date it'd need to be redone. 

We use the PR ourselves directly at https://docs.idex.market but for our next gen docs are switching to even more custom options such as docosaurus or gatsby, unfortunately.

RomanHotsiy:
@bradennapier sorry, we didn't have enough resources.

This is actually out of scope of this project. We try to keep it simple.

PR #855, opened 2019-03-19
Clean up how field values are displayed

drjonnicholson:
This PR fixes some minor presentation issues with enums etc.

### DIFF
--- a/src/components/Fields/EnumValues.tsx
+++ b/src/components/Fields/EnumValues.tsx
@@ -20,7 +20,7 @@ export class EnumValues extends React.PureComponent<EnumValuesProps> {
         </FieldLabel>{' '}
         {values.map((value, idx) => (
           <ExampleValue key={idx}>
-            {typeof value === 'string' ? value : JSON.stringify(value)}
+            {JSON.stringify(value)}
           </ExampleValue>
         ))}
       </div>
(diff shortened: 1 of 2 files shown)


RomanGotsiy:
Ah, actually for `enum` and `default` values I am intentionally persisting double quotes to clearly distinguish `string` type.

Just a few examples: 
- `200` (number) vs `"200"` (string)
- `true` (bool) vs `"true"` (string)

Does it make sense?

drjonnicholson:
I get you, you're making sure the type is a bit more explicit through the value. 

Personally I though that, since the `type` is displayed on the line above, it is clear enough without them. To me this made it cleaner / more readable, but honestly I'm fine if it stays as is (minus the trailing spaces on the enum values hah!)

RomanGotsiy:
> Personally I though that, since the type is displayed on the line above, it is clear enough without them. 

Yes, but sometimes it can get unclear. Check out your own enum example above and notice `null` value at the end.

> (minus the trailing spaces on the enum values hah!

drjonnicholson:
Fair point well made 👍 

I've reverted to the stringify method, I believe this has been reverted to be easy to read and should still resolve the spacing issue. I'll update the PR description accordingly.

drjonnicholson:
Original Description

---

This is a simple PR to clean up other field/enum values as done in #847.

Examples from RC4:

**Defaults**
redoc-defaults
with this PR becomes:
image


**Enums**
redoc-enums
with this PR becomes:
image


Pinging @RomanGotsiy as you helped with the other PR.

RomanGotsiy:
thanks!

PR #856, opened 2019-03-19
Display each security requirement on a new line

drjonnicholson:
This PR just makes security requirements block level elements so that they read nicer, particularly when there are several alternatives available.

Current display:
image

with this PR becomes:
image


drjonnicholson:
@RomanGotsiy, any thoughts on this one?

PR #869, opened 2019-03-26
Fix: Ignore empty x-tagGroups array

### LINKED ISSUE #868: Empty x-tagGroups array
If `x-tagGroups: []` empty array will be passed - there will be no tag groups displayed and all the Tags will be hidden too.
Screenshot from 2019-03-27 02-22-03

I think empty `x-tagGroups` array should behave in the same way as it's not exists, just display all the tags without any groupping. Otherwise it requires server side preprocessing to check existence of the empty array and its removal.

antonkomarev:
Fixes #868 

If `x-tagGroups: []` empty array is passed - tag groups should be ignored. Without this PR it hides all the tags.


RomanGotsiy:
Thanks!

PR #894, opened 2019-04-19
Add lineHeight config for headings

kkalavantavanich:
I was facing a problem that the heading looks kinda squashed because the `lineHeight` setting for headings are too small. This pull requests adds a theme configuration for the `lineHeight` of the headings.


RomanGotsiy:
Thanks!

PR #895, opened 2019-04-19
Remove querystring from exposed URL

antograssiot:
If ReDoc is initiated on a page containing query strings in the URL, the generated URL is misbehaving as shown on the following image

Capture d’écran 2019-04-19 à 15 32 02

Live example at https://demo.api-platform.com/?ui=re_doc&spec_version=3

After this change the exposed URL would be `https://demo.api-platform.com/books` as expected


antograssiot:
updated as requested I think @RomanGotsiy

RomanGotsiy:
Thanks!

PR #902, opened 2019-04-30
fix: update apiKey in to be titleize

### LINKED ISSUE #880: Capitalize first letter of the "in" value in apiKey Security scheme
This definition:
```yaml
  securitySchemes:
    Signature:
      description: foo
      name: Authorization
      type: apiKey
      in: header
```
Results in this:
image

`in: Header` instead of `in: header` solves the problem, but `Header` is an invalid value according to the spec.

cfabianski:
fixes #880


RomanGotsiy:
@cfabianski could you create a `titleize` function in `utils/helpers.ts` and use it in JSX?

cfabianski:
@RomanGotsiy Is that better?

RomanGotsiy:
Much better! Thanks!

PR #924, opened 2019-05-23
Non trivial enums need more documentation

stueynz:
[jsonSchema (draft 6)](https://json-schema.org/draft-06/json-schema-release-notes.html) defines '**const**' as a synonym with a single value '**enum**' 

So I've adjusted schema definition to do the right thing with const:

Also, the approved jsonSchema mechanism to document individual enum values is as follows:
(note the use of title and description tags to document the individual value meaning)

```
  Authority:
        description: >-
          <p>A code denoting the authority by which the Provider is authorised to provide education services. </p>
        oneOf:
          - const: "1"
            title: State
            description: >-
              "Also known as 'State:Non-integrated'"
          - const: "2"
            title: "State: Integrated"
          - const: "3"
            title: "Private: Fully Registered"
          - const: "4"
            title: "Private: Provisionally Registered"
          - const: "5"
            title: "Other: Vote Education"
            description: >-
              Owned by another organisation, funded from Vote Education.
          - const: "6"
            title: Partnership School
```
Which displays like this:
image

The display type of the field shouldn't aggregate all the details from all the variants in the oneOf - there might be too many.  So if the parent type is already defined, don't do the aggregation. 

So this input:

```
     Authority:
        description: >-
          <p>A code denoting the authority by which the Provider is authorised to provide education services. </p>
        type: string    # type is defined, so no aggregate type from the oneOf
        oneOf:
          - const: "1"
            title: State
            description: >-
              "Also known as 'State:Non-integrated'"
...
```

would look like this:

image

**Question:**  Maybe there should be an option or an x-something tag to drive this behaviour?


RomanHotsiy:
This PR adds too many changes and there is no clear use case for those changes.

I'm going to close this PR.
Please, open smaller PRs or open an issue first to discuss implementation.

Thanks!

PR #939, opened 2019-06-07
Added support for providing custom redoc JS source.

konradbartecki:
Ability to specify custom ReDoc source that will be served or bundled with  `redoc-cli (...) --customScript .\redoc.standalone.js` option.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -29,6 +29,7 @@ interface Options {
   templateFileName?: string;
   templateOptions?: any;
   redocOptions?: any;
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
This won't work as expected. The CLI uses Redoc lib to do Server Side rendering. If the code used to SSR differs from code loaded to the page the resulting page may be inconsistent or can even crash.

I'm going to close this PR. I see what's your use case though but I don't have any ideas of how to implement this reliable for now.

PR #958, opened 2019-06-28
Add missing content type header on compressed responses of /

deviantintegral:
I'm hosting an instance of redoc behind an HTTPs proxy at tugboat.qa. If a browser requests a compressed response, redoc-cli doesn't emit a `Content-Type` header. The HTTPs proxy is then inferring a `Content-Type` of `application/octet`, causing browsers to download the page instead of displaying it.

This PR sets the content type for `/`.


RomanHotsiy:
Thanks 👍

PR #963, opened 2019-07-04
update: fold active MenuItem if clicked

leoendless:
Fold and expand menu items by clicking


RomanHotsiy:
The previous behavior is intended and changing it by default is not good.

Add please an option for enabling this, e.g. `menuToggle`

RomanHotsiy:
Awesome! Thanks @leoendless!

PR #968, opened 2019-07-08
Add option for skipping enum quotes

stasiukanya:


src/services/RedocNormalizedOptions.ts
```diff
@@ -160,5 +162,6 @@ export class RedocNormalizedOptions {
     this.unstable_ignoreMimeParameters = argValueToBoolean(raw.unstable_ignoreMimeParameters);
 
     this.allowedMdComponents = raw.allowedMdComponents || {};
+    this.enumSkipQuotes = argValueToBoolean(raw.enumSkipQuotes);
```
RomanHotsiy: Move this to the big group above, after `hideSingleRequestSampleTab`

Issue #565, opened 2018-07-18
servers - variable substitutions not supported


fogine:
OAS: v3
the following configuration should be supported:  

```javascript
{
    servers: [
        {
             url: '{protocol}{host}{basePath}',
             variables: {
                 protocol: {default: 'http://'},
                 host: {default: '127.0.0.1'},
                 basePath: {default: ''}
        }
    ]
}

```

currently it renders endpoints as `{protocol}{host}{basePath}` instead of `http://127.0.0.1`

I might make a PR.

RomanGotsiy:
@fogine go for it 🏃!
Thanks!

PR #1006, opened 2019-08-07
Revert expanding default server variables

stasiukanya:
Servers are just displayed in Redoc and are not used to make calls so expanding variables may confuse you that the server doesn't have variables:

e.g.:

`https://{stage}.example.com` - makes it clear that the stage can be configurable

vs  `https://prod.example.com` - you have no way to learn about other stages and that they are configurable

### DIFF
--- a/src/components/Endpoint/Endpoint.tsx
+++ b/src/components/Endpoint/Endpoint.tsx
@@ -5,7 +5,7 @@ import { Markdown } from '../Markdown/Markdown';
 import { OptionsContext } from '../OptionsProvider';
 import { SelectOnClick } from '../SelectOnClick/SelectOnClick';
 
-import { expandDefaultServerVariables, getBasePath } from '../../utils';
+import { getBasePath } from '../../utils';
 import {
   EndpointInfo,
   HttpVerb,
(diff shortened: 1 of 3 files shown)


RomanHotsiy:
Agree.

Related to #565, cc @fogine

fogine:
Hm, I get your point and in terms of API documentation it might make more sense to not expand the variables. However in my opinion it depends on specific use case.

We are using `redoc` for rendering documentation for larger number of microservices and really do not need to show that the endpoint is configurable to the end user.

According to `OAS v3`:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject

one would expect that the variables will be substituted.

Maybe, this behavior could be configurable?

antherkiv:
I strong believe it's a mistake, a better way to handle this it's make the behavior optional, but not removed, In my case, the urls, became to mean nothing at all, because variables aren't displayed anyway. So, the pull request breaks the functionality replacing with nothing at all.

RomanHotsiy:
@fogine, @antherkiv could you elaborate on your use case. If you don't need to show that the endpoint is configurable why you use server variables at all? Why you just won't replace it in the server URL itself?

I don't mind making it configurable. Just want to understand.

antherkiv:
@RomanHotsiy, I'm using server variables, so now Redoc doesn't support the displaying of all of them, displaying at least the default grant us, to allow frontend developers to prepare the code in development for example at this moment and figure out how the api actually works.

I have done a "openapi collector" based on jsonschema of the apis, and several additional pieces, recollected along side the repo. So, I need to retain the server variable for the future.

Supporting the display of it, it's worth I guess, although the request inside the Redoc isn't possible and allow us to support the feature in the future easy.

I'm cloning the repo and I'll make a draft doing the feature optional.

antherkiv:
First at all, thanks because you have done a great job, here, I just love Redoc.

fogine:
@RomanHotsiy 
We use single source of truth for apis and although we might want to show placeholders for development team eg. as mentioned signalizing different environments being available, we dont want to show placeholders for APIs which are made publicly accessible - in this case there should be always explicit url available for the users.

It would be unfortunate to expect multiple versions of `api` specification to be generated with and without variable substitution just because of how `redoc` would render it.

antherkiv:
@RomanHotsiy , @fogine, I recovered the ability to expand default values on server variables, but it's optional now, for this people that wants to expands it for any reason. It's the pull request #1014.

Thanks and best regards.

antherkiv:
The main reason to support the rendering of default server variables values, it's to actually follows the [openapi spec](https://swagger.io/specification/#serverVariableObject), if not I guess we are not follow the spec, at least rendering the default it's expected I guess, if there are no value in enum. Best regards and I sorry for my insistence.

PR #1009, opened 2019-08-09
Correct typo of expolde to explode

AnthonyPorthouse:
This PR simply corrects a typo that happens when the console warns about deepObjects being used incorrectly


RomanHotsiy:
thanks!

PR #1011, opened 2019-08-13
request code samples could fetch the sample files

kylechenDEV:
**x-code-sample** sometimes is too much in the docs. The feature allows us to move sample files to the source with the prefix "@url:".
eg: 
before
```json
[ 
  { "lang": "Shell", "source": "curl -X GET http://api/samples/test " } ,
  { "lang": "Python", "source": "... " } 
]
```
after
```json
[ 
  { "lang": "Shell", "source": "@url:samplefiles/shell/samples/test.sh" } ,
  { "lang": "Python", "source": "@url:samplefiles/python/samples/test.py" } 
]
```

### DIFF
--- a/src/components/RequestSamples/RequestSamples.tsx
+++ b/src/components/RequestSamples/RequestSamples.tsx
@@ -1,6 +1,7 @@
 import { observer } from 'mobx-react';
 import * as React from 'react';
 import { OperationModel, RedocNormalizedOptions } from '../../services';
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
This is too specific and may be unpredictable. It would be simpler fro you to add a simple preprocessing script before passing a spec to Redoc.

Thanks for your PR but this is out of scope.

Issue #997, opened 2019-07-31
[#814] Unfixed problem, dependency 'core-js/es6/promise' is missing


eau-de-la-seine:
Hi, I still have the problem declared in the [issue number 814](https://github.com/Redocly/redoc/issues/814)

**Problem**:
```
Failed to compile
./node_modules/redoc/bundles/redoc.lib.js
Module not found: Can't resolve 'core-js/es6/promise' in '/home/gekinci/projects/pmu/pmu-redoc/node_modules/redoc/bundles'
```

**My dependencies** (package.json):
```
  "dependencies": {
    "core-js": "^3.1.4",
    "es6-promise": "^4.2.8",
    "mobx": "^4.13.0",
    "react": "^16.8.6",
    "react-dom": "^16.8.6",
    "react-scripts": "3.0.1",
    "redoc": "^2.0.0-rc.10",
    "styled-components": "^4.3.2"
  }
```

Installed with:
```
npm install mobx@^4.2.0 styled-components core-js es6-promise
npm install redoc --save
```

So I've installed the following dependency: `core-js` but nothing has changed. So I've looked into your Redoc code in `/node_modules/redoc/bundles/redoc.lib.js`:
```
module.exports = require("core-js/es6/promise");
```

But what you're looking for  with `core-js` is installed here: `core-js/es/promise`, not here: `core-js/es6/promise`

**Workaround**:
So my **workaround** was to copy/paste the existing `/node_modules/core-js/es` folder and rename it to `/node_modules/core-js/es6`. Is this a bug or did I miss something during the installation of Redoc with npm?

eau-de-la-seine:
I don't understand why you have closed this issue without indicating what the problem was, or what I've had to do to resolve it.
The best workaround I've found was to install this old version of `core-js`: **2.6.8** because it contains `es6` (also `es5` and `es7`) folders.

qw-in:
I updated `2.0.0-rc.12 => 2.0.0-rc.14` and ran into this issue. Reverting for now but +1

RomanHotsiy:
@eau-de-la-seine Just upgrading redoc to rc.14 and core-js to version 3 should work.

@qw-in 	could you provide more details? Have you upgraded core-js to version 3?

qw-in:
@RomanHotsiy will give that a shot, thanks

Maybe corejs@3 should be added [here](https://github.com/Redocly/redoc#usage-as-a-react-component)?

eau-de-la-seine:
[...]

PR #1018, opened 2019-08-16
Add core-js as peer dependency in docs

qw-in:
Since the redoc react component will not work without core-js, it should be added to the docs

Ref #997


transitive-bullshit:
Agreed; RedocStandalone doesn't work out of the box with `create-react-app` without this.

Issue #356, opened 2017-10-09
Slow load time due to fonts.gstatic.com


stianlagstad:
I'm experiencing very slow load times because some resources from fonts.gstatic.com take a long time to load. Take a look at this:
![screen shot 2017-10-09 at 11 16 52](https://user-images.githubusercontent.com/4340859/31331713-c24e5f2a-ace3-11e7-9871-8242d2272461.png)
Each of those resources that take 15 seconds are from fonts.gstatic.com. Should I have this problem?

RomanGotsiy:
Hey @stianlagstad,

I don't know why fonts.gstatic.com takes so much time for you. It is google fonts CDN.
Maybe something wrong with your network?

What `ping fonts.gstatic.com` returns?

stianlagstad:
Yeah it's weird. Do you know if there's a simple way to avoid fetching those fonts?

```
➜  swagger git:(master) ping fonts.gstatic.com
PING gstaticadssl.l.google.com (172.217.17.35): 56 data bytes
64 bytes from 172.217.17.35: icmp_seq=0 ttl=53 time=28.353 ms
64 bytes from 172.217.17.35: icmp_seq=1 ttl=53 time=27.658 ms
64 bytes from 172.217.17.35: icmp_seq=2 ttl=53 time=26.734 ms
64 bytes from 172.217.17.35: icmp_seq=3 ttl=53 time=27.005 ms
64 bytes from 172.217.17.35: icmp_seq=4 ttl=53 time=26.956 ms
64 bytes from 172.217.17.35: icmp_seq=5 ttl=53 time=26.947 ms
...
```

RomanGotsiy:
@stianlagstad, unfortunately, they are hardcoded in [here](https://github.com/Rebilly/ReDoc/blob/master/lib/components/Redoc/redoc-initial-styles.scss#L1-L2).

It will be fixed in the upcoming major release (React rewrite)

A quick workaround would be to download redoc.min.js bundle and find-and-replace those links.

Closing this issue, feel free to continue discussion here though

PR #1045, opened 2019-09-24
feat: add `disableGoogleFont` parameter to cli

duxiaofeng-github:
there is a hardcode in html template to force load a font family from google. same issue here: https://github.com/Redocly/redoc/issues/356

i dont think it's a good idea because of people in china has GFW there.

so i implement a switch called `disableGoogleFont` to turn it down and let people decide to open it or not. (it's still opening by default).


RomanHotsiy:
thanks!

PR #1058, opened 2019-10-09
Show examples in objects.

kimchisus:
I'd like Redoc to support examples inside ObjectSchema fields. 

image

### DIFF
--- a/src/components/Schema/ObjectSchema.tsx
+++ b/src/components/Schema/ObjectSchema.tsx
@@ -63,7 +63,7 @@ export class ObjectSchema extends React.Component<ObjectSchemaProps> {
                   undefined
                 }
                 className={field.expanded ? 'expanded' : undefined}
-                showExamples={false}
+                showExamples={true}
                 skipReadOnly={this.props.skipReadOnly}
                 skipWriteOnly={this.props.skipWriteOnly}
                 showTitle={this.props.showTitle}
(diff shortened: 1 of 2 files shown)


cviebrock:
@RomanHotsiy Any chance of having this reviewed and (hopefully) merged soon?  We're hoping to use it for a fairly large project here at Priceline and are trying to avoid forking.

Thanks!

huksley:
It would be good to have it merged, as we have schemas with examples!

TeemuKurki:
We also would like this to be merged. I think it is clearer and more intuitive to see example value directly on schema object than trying to find it on JSON example on the side, especially on larger schemas.

cah-alexsukhodolsky:
@RomanHotsiy can we please get this merged in? This is a very useful feature.

RomanHotsiy:
Thanks for the PR. I have the only request. Could you enable this functionality by top-level redoc option vs by default? Thanks!

TeemuKurki:
@RomanHotsiy As this PR seems to have died should I create a new PR with your requested features? Or is @rykimchi still on this?

jeremyfiel:
@RomanHotsiy   Can this be merged?
we have a very messy `br` html tag being used in the `description` to render examples.  each of them need to be hand written and they are often conflicting with the actual `example` value in the contract.

RomanHotsiy:
We'll have to finish one request first:

> Could you enable this functionality by top-level redoc option vs by default? Thanks!

This will be a breaking change if enable this behavior by default. I'll prioritize it.

PR #1059, opened 2019-10-10
Resolve formatting bugs.

russellrobinson:
**No quotes for default values in header fields.**
In a header field/parameter line Accept-Language, instead of 
`Default: "en-US"`
(which is ugly and just wrong), you now get:
`Default: en-US`

**Spaces around the pattern values.**
Instead of:
`string/^\d+$/`
you now get:
`string /^\d+$/`

Small README improvement.

### DIFF
--- a/demo/openapi.yaml
+++ b/demo/openapi.yaml
@@ -86,6 +86,14 @@ x-tagGroups:
 paths:
   /pet:
     parameters:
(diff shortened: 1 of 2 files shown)


adamaltman:
Thanks for your contribution @russellrobinson.

🎉 

It looks good to me.

PR #1064, opened 2019-10-12
wrap json examples in code tag

### LINKED ISSUE #578: Formatting of code is not wrapped in <code>
I tried adding the Google Translate widget to our references (I know that seems crazy, but it was a demand). The results were disastrous, so I went digging and discovered there are no &lt;codeGGT tags in the source -- all formatting is done with spans and many classes instead. Just curious as to why we're not just using code tags...

AndriyTykhan:
Fixes #578

### DIFF
--- a/src/utils/jsonToHtml.ts
+++ b/src/utils/jsonToHtml.ts
@@ -4,7 +4,7 @@ export function jsonToHTML(json, maxExpandLevel) {
   level = 1;
   let output = '';
   output += '<div class="redoc-json">';
-  output += '<code>'
+  output += '<code>';
   output += valueToHTML(json, maxExpandLevel);
   output += '</code>'
   output += '</div>';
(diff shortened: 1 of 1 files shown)


src/utils/jsonToHtml.ts
```diff
@@ -4,7 +4,9 @@ export function jsonToHTML(json, maxExpandLevel) {
   level = 1;
   let output = '';
   output += '<div class="redoc-json">';
+  output += '<code>'
   output += valueToHTML(json, maxExpandLevel);
+  output += '</code>'
```
RomanHotsiy:
```suggestion
  output += '</code>';
```

src/utils/jsonToHtml.ts
```diff
@@ -4,7 +4,9 @@ export function jsonToHTML(json, maxExpandLevel) {
   level = 1;
   let output = '';
   output += '<div class="redoc-json">';
+  output += '<code>'
```
RomanHotsiy:
```suggestion
  output += '<code>';
```

src/utils/jsonToHtml.ts
RomanHotsiy: @AndriyTykhan could you add semicolons?

RomanHotsiy:
Thanks :+1!

PR #1065, opened 2019-10-12
feat: display `multipleOf` constrains

### LINKED ISSUE #1055: How can I define how many digitis is allowed behind the comma?
How can I specify how many digitis is allowed for decimal part by using OAS 3.0/Redoc?

decimal(10,2) or NUMBER(10,2) means total length is 10 digits but out of 10 digits 2 digits are for decimal part. So the final allowed number will be 8 digits for integral part and 2 digits for fractional part.

nanov:
Fixes #1055

### DIFF
--- a/src/utils/__tests__/openapi.test.ts
+++ b/src/utils/__tests__/openapi.test.ts
@@ -362,7 +362,7 @@ describe('Utils', () => {
 
     it('should have a humanized constraint when multipleOf is set, and it is in format of /^0\\.0*1$/', () => {
       expect(humanizeConstraints(itemConstraintSchema(undefined, undefined, 0.01))).toContain(
-        'decimals <= 2 digits',
+        'decimal places <= 2',
       );
     });
 
--- a/src/utils/__tests__/openapi.test.ts
+++ b/src/utils/__tests__/openapi.test.ts
@@ -356,17 +356,17 @@ describe('Utils', () => {
       expect(humanizeConstraints(itemConstraintSchema(7, 7))).toContain('7 items');
     });
 
-    it('should have a humazined constraint when justMinItems is set, and it is equal to 1', () => {
+    it('should have a humanized constraint when justMinItems is set, and it is equal to 1', () => {
       expect(humanizeConstraints(itemConstraintSchema(1))).toContain('non-empty');
     });
 
-    it('should have a humazined constraint when multipleOf is set, and it is in format of /^0\\.0*1$/', () => {
+    it('should have a humanized constraint when multipleOf is set, and it is in format of /^0\\.0*1$/', () => {
       expect(humanizeConstraints(itemConstraintSchema(undefined, undefined, 0.01))).toContain(
         'decimals <= 2 digits',
       );
     });
 
-    it('should have a humazined constraint when multipleOf is set, and it is in format other than /^0\\.0*1$/', () => {
+    it('should have a humanized constraint when multipleOf is set, and it is in format other than /^0\\.0*1$/', () => {
       expect(humanizeConstraints(itemConstraintSchema(undefined, undefined, 0.5))).toContain(
         'multiple of 0.5',
       );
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/openapi.test.ts
```diff
@@ -358,6 +359,18 @@ describe('Utils', () => {
     it('should have a humazined constraint when justMinItems is set, and it is equal to 1', () => {
       expect(humanizeConstraints(itemConstraintSchema(1))).toContain('non-empty');
     });
+
+    it('should have a humazined constraint when multipleOf is set, and it is in format of /^0\\.0*1$/', () => {
```
adamaltman:
```suggestion
    it('should have a humanized constraint when multipleOf is set, and it is in format of /^0\\.0*1$/', () => {
```

src/utils/__tests__/openapi.test.ts
```diff
@@ -358,6 +359,18 @@ describe('Utils', () => {
...
+      expect(humanizeConstraints(itemConstraintSchema(undefined, undefined, 0.01))).toContain(
+        'decimals <= 2 digits',
+      );
+    });
+
+    it('should have a humazined constraint when multipleOf is set, and it is in format other than /^0\\.0*1$/', () => {
```
adamaltman:
```suggestion
    it('should have a humanized constraint when multipleOf is set, and it is in format other than /^0\\.0*1$/', () => {
```

nanov:
Typo is fixed, to the new tests as well as to the ```non-empty``` test.

src/utils/openapi.ts
```diff
@@ -368,6 +368,17 @@ export function isNamedDefinition(pointer?: string): boolean {
...
 
+function humanizeMultipleOfConstraint(multipleOf: number | undefined): string | undefined {
+  if (multipleOf === undefined) {
+    return;
+  }
+  const strigifiedMultipleOf = multipleOf.toString(10);
+  if (!/^0\.0*1$/.test(strigifiedMultipleOf)) {
+    return `multiple of ${multipleOf}`;
+  }
+  return `decimals <= ${strigifiedMultipleOf.split('.')[1].length} digits`;
```
RomanHotsiy: @nanov @adamaltman What do you think if we change it to `n decimals max` instead of `decimals <= n`?

nanov: Well, I have nothing against it.

The only thing i might thing of is that we may provide those hardcoded strings as configurations options to allow internationalisation.

nanov: having this in mind having strings as ```decimals <= {{x}}``` are some how easier to manage, translation wise speaking without introduction of (simple) templating system or a i18n solution.

adamaltman: These are both wrong actually.  I'm glad that you brought it up.

Decimals is the concept.  Decimal, from the dictionary:

> a fraction whose denominator is a power of ten and whose numerator is expressed by figures placed to the right of a decimal point.

What we're talking about here specifically is called decimal places. 

So it should be like this:

* decimal places <= n
* n decimal places max

However, in the second phrase, `n decimal places max` the places must turn singular if `n == 1`.

nanov: If we decide to go with the second option we could go with the good old:

* n decimal place(s) max

As some languages have also a double form ( ie. singular, double, and plural ) i think it is better to use simpler descriptions in order to make internationalization simpler, with minimal overhead when the time comes.

nanov: I've taken your first suggestion and implemented it, I took the time to see how much effort would it be to implement mini translating system, and it seems as a straight forward task ( can work similar to the theme option ) and for our use case it can be done with zero-dependencies.

The question here is should i implement it here, or do a new PR for it?

nanov:
Can we somehow proceed with this, rather simple, PR, i would like to (try) and implement a internationalisation system, but as this PR is tightly connected to it I would prefer it gets merged first.

nanov:
Any updates on this one?

RomanHotsiy:
@nanov thanks a log and sorry for the delay.

Issue #896, opened 2019-04-23
Link to sections sometimes broken on Firefox


karussell:
Sometimes simple links do not work. E.g.: https://docs.graphhopper.com/#operation/solveVRP gives this wrong result on Firefox:

![image](https://user-images.githubusercontent.com/129644/56519424-4f0cda00-6542-11e9-9cab-485deb2ce858.png)

(Chrome is ok)

This issue is partially related to #63

RomanGotsiy:
I can verify that I can reproduce the issue.

RomanGotsiy:
Looks like it is caused by images loading (from the table above) which causes scroll event in FF and works fine in Chrome.

Don't know how to fix it yet. Just letting you know of the main reason. Maybe you will find a way to workaround for now.

adamaltman:
I can't reproduce this one.  Maybe a newer version of FF? I'm on 68.0.2 (64-bit)?

karussell:
> I can't reproduce this one. Maybe a newer version of FF? I'm on 68.0.2 (64-bit)?

We have fixed this in the linked repository. Additionally to some other stuff we will contribute this in some separate PRs.

RomanHotsiy:
@karussell Let us know what the fix was or contribute as a PR. Thanks!

karussell:
See now https://github.com/Redocly/redoc/pull/1067

karussell:
@RomanHotsiy Can you have a look into #1067 and if there are blockers for merging?

Issue #562, opened 2018-07-16
Search content keyword not bring into view


chapelloise:
Search a keyword which in the right side main content via the upper left search box. 
Let's take Rebilly as an example.(https://rebilly.github.io/RebillyAPI/) Search for "customerId", the word is found on several places, but they are not brought into view. Need to manually scroll down to see them as highlighted.

arnonzooz:
Hi. I experience the same. Moreover, collapsable nodes are not expanded if includes a found item. The overall experience using ReDoc is great, but the search functionality seems to be somewhat limited.

karussell:
We developed fixes for this issue (try [here](https://docs.graphhopper.com/) e.g. type "avoid"). The improved search also requires to point via links deeper into the document. I think these changes would be also of interests for others. Unfortunately only a few PRs seem to be merged. @RomanHotsiy would you have time to take a look into the first required PR #979 from @thisismo?

lornajane:
This should be fixed in newer versions of Redoc. I'm closing the issue, but please don't hesitate to let us know if you're still seeing a problem - it would be best to open a new issue with the version you're using and any other information you think we need.

PR #1067, opened 2019-10-14
Added Content-Fields to Search Index and fixed Firefox Link Bug

thisismo:
**Added each field to search index as requested in #562** 
Also, creates unique IDs for each field to allow jumping to them by individual URLs.

**Fixed links to properly show the right section in Firefox (#896)**

**PS:** The failed Travis Build was expected, as one of the unit tests compares the number of results for a certain char in the input with a static value. Since this PR improved the search, the number of results per char exceeds the static value and led to a failure of this unit test.


RomanHotsiy:
This is out of scope of this project. We keep search simple here on purpose.

Thanks for your efforts. I have to close it. Sorry for huge delay 🙈

monica-m-ps:
Hi @RomanHotsiy, Is there a reason there's no fix to this? I feel it's annoying when the search bar says "No results found" to the left but it actually highlights that word to the right and we have to scroll down to find that word. I am a technical writer and I'm very much interested in finding content in the API docs to speed up my tasks. Thanks

Issue #1072, opened 2019-10-17
Lint and prettify cli/index.ts


ZakCodes:
Currently, the configuration in the `package.json` for the `prettifier` script doesn't include the `cli/index.ts` file. It's the same thing for the `.tsconfig.json` file.  
Therefore, the `cli/index.ts` isn't prettified or linted.

PR #1073, opened 2019-10-17
feat(cli): Fallback on the spec's title before falling back on "ReDoc documentation"

### LINKED ISSUE #1068: Use the OpenAPI's title for rendered HTML pages
When using the CLI, if you don't explicitly provide a title using the `--title` option, the title of the outputed page is `ReDoc Documentation`. 

Instead of using the default title, ReDoc could use the title provided by the OpenAPI document that is being rendered. In the OpenAPI standard, there's an info object in the root of the document with a `title` property that could be used. This property is already used by ReDoc cli at the very beginning of the rendered HTML page.

I would understand if you would like to keep it like that for advertising, but I think it's really annoying because I need to copy the title from my OpenAPI document and pass it to ReDoc cli, so whenever I want to change a title, I need to remember to change it in my rendering scripts.  
I think the title of an HTML document isn't what's going to give you the most publicity since the users can't even click on it. Therefore, if you really want to advertise this tool, you would be better off adding a link to ReDoc right under the title in the HTML body where people can click on, just like you did in the left sidebar.

ZakCodes:
Instead of using "ReDoc documentation" straight away as the title of the HTML page rendered by `redoc-cli`, it tries to use the specification's title in its info object.

I haven't seen any place to write tests for the cli tool, so I didn't write any.  
However, here are the three possible test cases and expected results: 

#### Title provided using `--title`
Command: `redoc-cli bundle spec.yaml --title "Your Awesome Title"`
OpenAPI spec title: "Another title" | null
HTML page title (result): "Your Awesome Title"

#### Title provided in the specification file
Command: `redoc-cli bundle spec.yaml`
OpenAPI spec title: "Your Awesome Title"
HTML page title (result): "Your Awesome Title"

#### No title provided
Command: `redoc-cli bundle spec.yaml`
OpenAPI spec title: null
HTML page title (result): "ReDoc documentation"

I've still made sure that everything was prettified and linted which made me discover the issue #1072.

Fixes #1068

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -99,7 +99,6 @@ YargsParser.command(
       yargs.options('title', {
         describe: 'Page Title',
         type: 'string',
-        default: 'ReDoc documentation',
       });
 
       yargs.options('disableGoogleFont', {
@@ -291,7 +290,7 @@ async function getPageHTML(
         ? '<script src="https://unpkg.com/redoc@next/bundles/redoc.standalone.js"></script>'
         : `<script>${redocStandaloneSrc}</script>`) + css
       : '<script src="redoc.standalone.js"></script>',
-    title,
+    title: title || spec.info.title || 'ReDoc documentation',
     disableGoogleFont,
     templateOptions,
   });


RomanHotsiy:
Thanks!

StephaneLambert:
Thanks for that very usefull feature!
But IMHO, it doesn't seems to work for `serve` with redoc-cli 0.9.5.

The command: `redoc-cli serve spec.yaml --title "Your Awesome Title"`
=> Always show OpenAPI spec title instead of --title text.

ZakCodes:
@StephaneLambert I've tested what you've said with the latest version (v0.9.5) and you're right. Sorry for the inconvenience. I'll fix it this weekend for sure.

ZakCodes:
It turns out that `redoc-cli serve spec.yaml --title "Your Awesome Title"` never worked. You can easily see this just by running `redoc-cli serve --help` and `redoc-cli bundle --help`. You'll see that the first one doesn't show the option `--title` and the second one does.

PR #1077, opened 2019-10-18
Fixed example slack api url

alexgmin:
Since this seems like something that happens once in a while, is there a reason to not use a permalink, like linking to the yaml file in the apis.guru repo?
https://raw.githubusercontent.com/APIs-guru/openapi-directory/c8d61eeb3c52504cbe006838093897df20682342/APIs/slack.com/1.2.0/swagger.yaml


RomanHotsiy:
thanks

PR #1089, opened 2019-10-29
fix: Add debounce for 300 ms when searching

### LINKED ISSUE #1027: Searching bogs down while typing a search term
Reproducing:

1) Use redoc on a large spec.

2) Quickly type a phrase into the search bar

Result:

The UI hangs, a search is performed for each change event.

Fix:

Debounce searches, so a search is only performed when a certain amount of time has passed since the last change event.

faheem00:
Fixes #1027 

Steps taken
- Ran tslint
- Ran all tests
- Ran `yarn prettier`

### DIFF
--- a/src/components/SearchBox/SearchBox.tsx
+++ b/src/components/SearchBox/SearchBox.tsx
@@ -95,13 +95,12 @@ export class SearchBox extends React.PureComponent<SearchBoxProps, SearchBoxStat
   setResults(results: SearchResult[], term: string) {
     this.setState({
       results,
-      term,
     });
     this.props.marker.mark(term);
   }
 
   @bind
-  @debounce(300)
+  @debounce(400)
   searchCallback(searchTerm: string) {
     this.props.search.search(searchTerm).then(res => {
       this.setResults(res, searchTerm);
--- a/src/components/SearchBox/SearchBox.tsx
+++ b/src/components/SearchBox/SearchBox.tsx
@@ -7,6 +7,7 @@ import { MenuItem } from '../SideMenu/MenuItem';
 import { MarkerService } from '../../services/MarkerService';
 import { SearchResult } from '../../services/SearchWorker.worker';
 
+import { bind, debounce } from 'decko';
 import { PerfectScrollbarWrap } from '../../common-elements/perfect-scrollbar';
 import {
   ClearIcon,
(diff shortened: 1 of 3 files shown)


faheem00:
`Search should support arrow navigation` test seems to be failing. Is it because the search result function was debounced, and the result display is delayed?

src/components/SearchBox/SearchBox.tsx
```diff
@@ -106,13 +107,25 @@ export class SearchBox extends React.PureComponent<SearchBoxProps, SearchBoxStat
...
+          }),
+        300,
+      ),
+    );
+  };
 
-    this.props.search.search(event.target.value).then(res => {
-      this.setResults(res, q);
-    });
+  debounce = (fn: (..._) => any, wait: number = 1): (() => void) => {
```
RomanHotsiy: Could you remove your hand-made debounce and use `debounce` from `decko` (we already use this dependency) instead?

faheem00: I tried using the `debounce` implementation from decko as such - 
```javascript
@debounce
search = (event: React.ChangeEvent<HTMLInputElement>) => {
...
}
```
But this showed the following warning when trying to compile- 
`TS1240: The return type of a property decorator function must be either 'void' or 'any'. `

Furthermore, when I try to use `delay` as argument as such - 
```javascript
@debounce(300)
search = (event: React.ChangeEvent<HTMLInputElement>) => {
...
}
```
I get the following warning - 
`TS1240: 
Unable to resolve signature of property decorator when called as an expression.`

As the script was not compiling, I opted to use a hand-made debounce. One thing to note is that the files in which the `debounce` from `decko` is already being used (e.g. `ScrollService.ts` and `HistoryService.ts` are `ts` file, whereas `SearchBox` is a `tsx` file. Maybe there needs to be a change in lint configuration for `tsx` files?

RomanHotsiy: Use 

```
  @bind
  @debounce
  search(event: React.ChangeEvent<HTMLInputElement>) {
```

Note `search` is not field with lambda function as value but method.

faheem00: Pushed the changes

fabienpomerol:
@faheem00  guys any update on this subject ? i'm also interested by the fix :)

PR #1123, opened 2019-12-03
Support for required additionalProperties

### LINKED ISSUE #1086: Support for required additionalProperties in schema
When specifying `additionalProperties` fields as `required` there's no indication of these.

Example schema:
```yaml
openapi: 3.0.2
info:
  title: Test
  version: "0.0.1"
paths:
  /example:
    get:
      summary: Example operation
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                allOf:
                  - type: object
                    description: dynamic fields, expect to see `prop1` and `prop2` as required field names here
                    additionalProperties:
                      type: string
                  - required:
                    - prop1
                    - prop2
              example:
                prop1: required prop
                prop2: required prop
                prop3: optional prop
```

Vasfed:
This fixes #1086 

When resulting schema contains `required` fields, that are not explicitly listed in schema - they must be covered by `additionalProperties` (because openapi does not support `patternProperties`) schema (if any) or are just `{}`.

The point is in listing these names, because validation will fail without them

### DIFF
--- a/demo/openapi.yaml
+++ b/demo/openapi.yaml
@@ -398,10 +398,16 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                additionalProperties:
-                  type: integer
-                  format: int32
+                allOf:
+                  - type: object
+                    additionalProperties:
+                      x-additionalPropertiesName: status
+                      type: integer
+                      format: int32
+                  - required:
+                    - inStock
+                    - soldOut
+
       security:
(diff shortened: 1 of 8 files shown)


lornajane:
Thank you @Vasfed for sending us this. It's been such a long time, I'm going to admit that we're not going to merge this change and instead I'm closing this pull request.

PR #1126, opened 2019-12-05
Change the title of "Security Scheme Type" to match "HTTP Authorizati…

torbenw:
…on Scheme"

The way "Security Scheme Type" was written did not match the way "HTTP Authorization Scheme" is written in the security details table.


RomanHotsiy:
Makes sense 👍

Issue #1121, opened 2019-12-02
Mixed up descriptions after hydration of docs generated by the CLI


aryehb:
I'm having an issue with the CLI, where when the schema get re-hydrated, the field details get mixed up.

When I use the CLI to generate documentation with the schema below, the initial generated docs are fine. However, after they are re-hydrated, the descriptions will be mixed up, and I'll see `Field 6` as the description for `field1`. This only happens when there are at least 11 fields, and when using the `--options.requiredPropsFirst` flag, regardless of whether there are required fields in the schema.

```yaml
openapi: 3.0.0
info:
  title: API
  version: 1.0.0
paths:
  /:
    post:
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                field1:
                  type: string
                  description: Field 1.
                field2:
                  type: string
                  description: Field 2.
                field3:
                  type: string
                  description: Field 3.
                field4:
                  type: string
                  description: Field 4.
                field5:
                  type: string
                  description: Field 5.
                field6:
                  type: string
                  description: Field 6.
                field7:
                  type: string
                  description: Field 7.
                field8:
                  type: string
                  description: Field 8.
                field9:
                  type: string
                  description: Field 9.
                field10:
                  type: string
                  description: Field 10.
                field11:
                  type: string
                  description: Field 11.
```

Sample screenshot:

![image](https://user-images.githubusercontent.com/49538279/69981980-657c0a80-1501-11ea-840b-d7e0e24597c9.png)

fabong:
We are having the same issue. Also, it seems that this issue could be linked: https://github.com/Redocly/redoc/issues/1061.

aryehb:
Good point. I didn't see that issue.

fabong:
I just tested this PR: https://github.com/Redocly/redoc/pull/1107 and it fixes our issue.

yannan:
I have the same issue. So I pointed this bug in #1104 and try to fix it in #1107

Issue #1104, opened 2019-11-14
The method named `sortByRequired` of `src/utils/openapi.ts`  is not reliable.


yannan:
```
export function sortByRequired(
  fields: Array<{ required: boolean; name: string }>,
  order: string[] = [],
) {
  fields.sort((a, b) => {
    if (!a.required && b.required) {
      return 1;
    } else if (a.required && !b.required) {
      return -1;
    } else if (a.required && b.required) {
      return order.indexOf(a.name) - order.indexOf(b.name);
    } else {
      return 0;
    }
  });
}
```
See the usage of `sort` from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)
> If compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements. Note: the ECMAscript standard does not guarantee this behavior, thus, not all browsers (e.g. Mozilla versions dating back to at least 2003) respect this.

It may be cause the render result of ssr is not same between node endpoint and browser endpoint.Then the schema display unexpected when use ReactDOM hydrate.

yannan:
Open another PR

PR #1136, opened 2019-12-10
fix: Fixed  method sortByRequired

yannan:
The method of sortByRequired has a bug, see #1104, #1121

### DIFF
--- a/src/services/models/Operation.ts
+++ b/src/services/models/Operation.ts
@@ -154,12 +154,11 @@ export class OperationModel implements IMenuItem {
     ).map(paramOrRef => new FieldModel(this.parser, paramOrRef, this.pointer, this.options));
 
     if (this.options.sortPropsAlphabetically) {
-      sortByField(_parameters, 'name');
+      return sortByField(_parameters, 'name');
     }
     if (this.options.requiredPropsFirst) {
-      sortByRequired(_parameters);
+      return sortByRequired(_parameters);
     }
-    return _parameters;
   }
 
   @memoize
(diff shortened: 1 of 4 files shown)


RomanHotsiy:
Hey @yannan,

The `sortByReqired` method is not obvious to read... I tried to simplify the flow and managed to get tests passing.

Could you please replace it with the following and verify you don't have an issue with openapi-cli with this version:

```ts
export function sortByRequired(fields: FieldModel[], order: string[] = []) {
  return [...fields].sort((a, b) => {
    if (!a.required && b.required) {
      return 1;
    } else if (a.required && !b.required) {
      return -1;
    } else if (a.required && b.required) {
      const aIdx = order.indexOf(a.name);
      const bIdx = order.indexOf(b.name);
      if (aIdx === -1 && bIdx !== -1) {
        return 1;
      }
      if (bIdx === -1 && aIdx !== -1) {
        return -1;
      }

      return aIdx - bIdx;
    } else {
      return 0;
    }
  });
}
```

RomanHotsiy:
If it is, update this PR pls. I can't push to your branch

yannan:
@RomanHotsiy. Thank you for your reply.
But We have a issue when the compareFunction(a,b) return 0 in the end of if/else condition.

If compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements.the ECMAscript standard does not guarantee this behavior.

RomanHotsiy:
I see. It is recent addition: https://v8.dev/features/stable-sort

What about this approach: https://stackoverflow.com/a/52771420/1749888?

yannan:
Hey @RomanHotsiy.
I update this PR, please check.

PR #1185, opened 2020-02-26
fix: broken sticky sidebar

boojongmin:
I have trouble that hiding menu bar sometimes.
It is css issue. and I add 2 class in template file.

image

thank you.


RomanHotsiy:
this issue has been fixed in rc.23

PR #1186, opened 2020-02-27
Fix #1154 (falsy recursive with allOf)

### LINKED ISSUE #1154: False positive recursion
Hello there, wrestling following issue.

Imagine I want to reuse some schema, let's say ID – I have following in `./helpers/id.yaml`:

```yaml
properties:
  id:
    type: number
required:
  - id
```

And here is a simple OpenAPI, which should result in a simple structure of three nested objects, which all wants to reference the `id` helper.

```yaml
openapi: 3.0.0
info:
  title: False positive recursion
  version: '3.0'
paths:
  /documents:
    get:
      summary: Example
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                allOf:
                  - $ref: ./helpers/id.yaml
                  - properties:
                      first:
                        properties:
                          second:
                            allOf:
                              - $ref: ./helpers/id.yaml
                              - properties:
                                  third:
                                    allOf:
                                      - $ref: ./helpers/id.yaml
```

Whereas the first and second object are correctly rendered with required `id` attribute, the third object is flagged as recursive:

image

Any idea? Is it a bug or am I missing something?

m0003r:
Now it immediately exits from allOf's children, but use separate refCounter to prevent entering into same scheme recursively

Also I have added pseudo-recursive example from #1154

### DIFF
--- a/demo/pseudo-recursive.yaml
+++ b/demo/pseudo-recursive.yaml
@@ -1,0 +1,58 @@
+openapi: 3.0.0
+info:
+  title: False positive recursion
+  version: '3.0'
+tags:
+  - name: a
+
+components:
+  schemas:
+    id:
+      tags:
+        - a
+      type: object
+      properties:
+        id:
+          type: number
+          description: id
+    pet:
+      tags:
+        - a
+      type: object
+      properties:
+        name:
+          type: string
+        friends:
+          type: array
+          items:
+            allOf:
+              - $ref: '#/components/schemas/pet'
+paths:
+  /documents:
+    get:
+      summary: Example
+      responses:
+        '200':
+          description: Pseudo-recursvie
+          content:
+            application/json:
+              schema:
+                allOf:
+                  - $ref: '#/components/schemas/id'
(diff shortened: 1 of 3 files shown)


krystof-k:
Thanks for the PR. I tried it but it's still not working with my example from #1154 – or I'm doing something wrong.

m-mohr:
Any news on this? Would be great to get a fix for this issue.

andy-williams:
> Thanks for the PR. I tried it but it's still not working with my example from #1154 – or I'm doing something wrong.

Just tested this with the `.yaml` file provided and it seems to work ok 👍

krystof-k:
You're right, I was probably trying out wrong version. Let's merge it :)

alanag13:
@RomanHotsiy Any hope of this getting merged soon? It resolves my issue and I'd rather not have to resort to installing redoc by building a fork.

RomanHotsiy:
We have resolved it in https://github.com/Redocly/redoc/pull/2072

Issue #764, opened 2018-12-28
Content Security Policy error on redoc


unional:
When I run a page on php using redoc@2.0.0-rc.0, It errors with the following:

```
Something went wrong...
Failed to construct 'Worker': Access to the script at 'blob:http://<host>/53f69b15-64aa-4228-ab20-4271b18d44f8' is denied by the document's Content Security Policy.
Stack trace
Error: Failed to construct 'Worker': Access to the script at 'blob:http://<host>/53f69b15-64aa-4228-ab20-4271b18d44f8' is denied by the document's Content Security Policy.
    at new e.exports (http://<host>/doc/redoc.standalone.js:84:98978)
    at new e (http://<host>/doc/redoc.standalone.js:107:56868)
    at new e (http://<host>/doc/redoc.standalone.js:113:26422)
    at t.makeStore (http://<host>/doc/redoc.standalone.js:107:60569)
    at t.render (http://<host>/doc/redoc.standalone.js:107:61475)
    at ci (http://<host>/doc/redoc.standalone.js:68:59752)
    at li (http://<host>/doc/redoc.standalone.js:68:59547)
    at di (http://<host>/doc/redoc.standalone.js:68:63224)
    at Vi (http://<host>/doc/redoc.standalone.js:68:78336)
    at Qi (http://<host>/doc/redoc.standalone.js:68:78713)

ReDoc Version: 2.0.0-rc.0 
Commit: a47a4b8
```

The problem is caused by this:
```js
var e=new Worker(URL.createObjectURL(new Blob(['/*!\n * ReDoc - OpenAPI/Swagger-generated API Reference...
```
~Source:~

This might be a security concern that should be addressed.

EDIT: source is incorrect. Looking up the right code.

unional:
It seems to be introduced by calling `webpack.BannerPlugin()` in the webpack.config.ts.

unional:
There is another one, but it seems like just bundling `lunr`. scratching my head over this one.

```js
[...]

PR #1191, opened 2020-02-29
fix: do not load SearchWorker if disableSearch is `true`

PixnBits:
Blob URLs may be disabled by CORS (ex: #764) but should not block execution when the disableSearch option is true


RomanHotsiy:
Thanks!

PR #1198, opened 2020-03-05
[redoc-cli] Enables url watching.

ryan-sandy:
Currently specs are only fetched one from a url on start.

This PR allows the `--watch` option to fetch URLs on each request.

`redoc-cli --watch server http://localhost/api/doc`


ryan-sandy:
I am not super familiar with Typescript, or this library, so any feedback would be appreciated.

cli/index.ts
```diff
@@ -159,8 +161,16 @@ YargsParser.command(
 async function serve(port: number, pathToSpec: string, options: Options = {}) {
   let spec = await loadAndBundleSpec(pathToSpec);
   let pageHTML = await getPageHTML(spec, pathToSpec, options);
+  let isUrl;
+  try {
+    // tslint:disable-next-line
+    new URL(pathToSpec);
```
RomanHotsiy, line 167: This is not a valid URL check, e.g. `./some/folder/openapi.js` is a valid relative URL. You have to check if this is an absolute URL. I would keep it simple, just check if `pathToSpec` starts with `http://` or `https://`

cli/index.ts
```diff
@@ -176,6 +186,12 @@ async function serve(port: number, pathToSpec: string, options: Options = {}) {
         'Content-Type': 'text/html',
       });
     } else if (request.url === '/spec.json') {
+      if (options.watch && isUrl) {
+        spec = await loadAndBundleSpec(pathToSpec).catch(e => {
```
RomanHotsiy, line 190: You don't need `.catch` when you do `await`. Any error will be thrown and can be catched by try/catch.

RomanHotsiy:
Thanks for your PR, I left you a few change requests.

RomanHotsiy:
We deprecated serve. Use openapi-cli preview-docs for the same functionality: https://redocly.com/docs/cli/commands/preview-docs/

PR #774, opened 2019-01-08
Show all the mapping entries

zeapo:
When the discriminator has `mapping` with more than one key pointing to the same ref. Only the last one was shown.
The following example would show only `kitten`, `dog` and `bee` in the dropdown. We'd expect to have all four (i.e. `cat` also!).
```yaml
      discriminator:
        propertyName: petType
        mapping:
          cat: '#/components/schemas/Cat'
          kitten: '#/components/schemas/Cat'
          dog: '#/components/schemas/Dog'
          bee: '#/components/schemas/HoneyBee'
```
In this PR. The dropdown will show all the mappings:
dropdown_1

Another use-case is when the mapping is partial, here the `dog` mapping was removed:
```yaml
      discriminator:
        propertyName: petType
        mapping:
          cat: '#/components/schemas/Cat'
          kitten: '#/components/schemas/Cat'
          bee: '#/components/schemas/HoneyBee'
```
The type `Dog` will be included in addition to the mappings:
dropdown_2

The third use-case is an exhaustive mapping, in that case we'd like to ignore non-mapping child classes. I've added a new extension `x-limitToMapping` of type boolean. When used, it excludes the class `Dog` in the following case:
```yaml
      discriminator:
        propertyName: petType
        x-limitToMapping: true
        mapping:
          cat: '#/components/schemas/Cat'
          kitten: '#/components/schemas/Cat'
          bee: '#/components/schemas/HoneyBee'
```
will yield the following result:
dropdown_3

PR #1215, opened 2020-03-26
feat: add x-explicitMappingOnly extension

zeapo:
This replaces #774 

When the mapping is partial, here the `dog` mapping was removed:
```yaml
      discriminator:
        propertyName: petType
        mapping:
          cat: '#/components/schemas/Cat'
          kitten: '#/components/schemas/Cat'
          bee: '#/components/schemas/HoneyBee'
```
The type `Dog` will be included in addition to the mappings:
dropdown_2

The idea is to have an exhaustive mapping, in that case we'd like to ignore non-mapping child classes. I've added a new extension `x-limitToMapping` of type boolean. When used, it excludes the class `Dog` in the following case:
```yaml
      discriminator:
        propertyName: petType
        x-limitToMapping: true
        mapping:
          cat: '#/components/schemas/Cat'
          kitten: '#/components/schemas/Cat'
          bee: '#/components/schemas/HoneyBee'
```
will yield the following result:
dropdown_3

### DIFF
--- a/docs/redoc-vendor-extensions.md
+++ b/docs/redoc-vendor-extensions.md
@@ -306,3 +306,37 @@ Player:
     x-additionalPropertiesName: attribute-name
     type: string
 ```
+
+#### x-explicitMappingOnly
+**ATTENTION**: This is ReDoc-specific vendor extension. It won't be supported by other tools.
+
+Extends the `discriminator` property of the schema object.
+
+| Field Name     |	Type	  | Description |
+| :------------- | :------: | :---------- |
+| x-explicitMappingOnly | boolean | limit the discriminator selectpicker to the explicit mappings only |
+
+###### Usage in ReDoc
+ReDoc uses this extension to filter the `discriminator` mappings shown in the selectpicker.
+When set to `true`, the selectpicker will only list the the explicitly defined mappings. When `false`,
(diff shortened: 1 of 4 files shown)


adamaltman:
Thanks for rewriting this clearly and referencing the original issue. :+1: :+1:

RomanHotsiy:
Ooh! @zeapo sorry for not responding. Tons of work here.

This PR is so much shorter 😍! Good job! 

I'm not just fully sold on the name `x-limitToMapping`

@adamaltman do you have any other ideas? @zeapo maybe you have some alternative?

zeapo:
Thanks :)

Maybe `x-explicitMappingOnly`? As you've named them explicit/implicit.

RomanHotsiy:
Well, it is how they call it in the spec:

image

I like `x-explicitMappingOnly` more. Could you also add some short docs to https://github.com/Redocly/redoc/blob/master/docs/redoc-vendor-extensions.md

As soon as you're done I am ready to merge this 👍

zeapo:
@RomanHotsiy , It's done. Let me know if anything is missing.

adamaltman:
I think this is a good addition.

RomanHotsiy:
@zeapo awesome! Thanks 🙇

Issue #1111, opened 2019-11-21
discriminator mapping 1-n


mohdzohlof:
When 2 or more values in the discriminator reference the same schema, only the last 1 is displayed (seems like the others get overwritten)

![Screenshot from 2019-11-21 15-47-58](https://user-images.githubusercontent.com/26624901/69343467-592cbd80-0c76-11ea-9080-cf5064b2bfe7.png)

```yaml
openapi: 3.0.0

########## INFO ##########
info:
  title: Test
  description: Specifications of test bounded context
  version: 1.0.0

paths:
  /pet:
    put:
      tags:
        - Pet
      description: Test
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pet"
        description: Capture that has to be applied to transaction
      responses:
        "204":
          description: The transaction captured.
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
      discriminator:
        propertyName: pet_type
        mapping:
          dog: "#/components/schemas/Dog"
          cachorro: "#/components/schemas/Dog"
          cat: "#/components/schemas/Cat"
          lizzard: "#/components/schemas/Lizard"
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Cat`
          properties:
            name:
              type: string
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Dog`
          properties:
            bark:
              type: string
    Lizard:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Lizard`
          properties:
            lovesRocks:
              type: boolean
```

SensibleWood:
+1

SensibleWood:
+1

PR #1216, opened 2020-03-26
fix: sort discriminator entries by mapping order

zeapo:
The fix for #1111 (in https://github.com/Redocly/redoc/commit/6e390f9c7909da0b5d1d6fc571ab4ad92e715d6e) sadly broke the order of mapping entries.
Given the discriminator:
```yaml
      discriminator:
        propertyName: petType
        x-limitToMapping: true
        mapping:
          dog: "#/components/schemas/Dog"
          cat: "#/components/schemas/Cat"
          doggy: "#/components/schemas/Dog"
          bee: "#/components/schemas/HoneyBee
```
The result is out of order:
image
Also, if we remove the `Dog` mappings, we end up with the `Dog` class between the mappings:
image

This PR adds a sort in case of the presence of a mapping. The behavior is quite simple:
- Explicit mappings are prioritized to implicit mappings
  ```yaml
      discriminator:
        propertyName: petType
        x-limitToMapping: true
        mapping:
          dog: "#/components/schemas/Dog"
          cat: "#/components/schemas/Cat"
          doggy: "#/components/schemas/Dog"
          bee: "#/components/schemas/HoneyBee"
  ```
  gives
  image

- Explicit mappings are sorted by the order they were defined in the mapping
- Implicit mappings are sorted alphabetically among themselves
  ```yaml
      discriminator:
        propertyName: petType
        x-limitToMapping: true
        mapping:
          cat: "#/components/schemas/Cat"
  ```
  image 
In case no mapping is defined, the default behavior is kept.

### DIFF
--- a/src/services/models/Schema.ts
+++ b/src/services/models/Schema.ts
@@ -283,7 +283,7 @@ export class SchemaModel {
 
         if (indexLeft < 0 && indexRight < 0) {
           // out of mapping, order by name
-          return left.name - right.name;
+          return left.name.localCompare(right.name);
         } else if (indexLeft < 0) {
           // the right is found, so mapping wins
           return 1;
--- a/src/services/models/Schema.ts
+++ b/src/services/models/Schema.ts
@@ -254,7 +254,7 @@ export class SchemaModel {
 
     const inversedMapping = { ...implicitInversedMapping, ...explicitInversedMapping };
 
-    const refs: Array<{ $ref; name }> = [];
+    let refs: Array<{ $ref; name }> = [];
 
     for (const $ref of Object.keys(inversedMapping)) {
       const names = inversedMapping[$ref];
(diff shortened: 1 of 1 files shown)


src/services/models/Schema.ts
```diff
@@ -267,6 +267,35 @@ export class SchemaModel {
...
+    // - Names outside the mapping are sorted alphabetically
+    const names = Object.keys(mapping);
+    if (names.length !== 0) {
+      refs = refs.sort((left, right) => {
+        const indexLeft = names.indexOf(left.name);
+        const indexRight = names.indexOf(right.name);
+
+        if (indexLeft < 0 && indexRight < 0) {
+          // out of mapping, order by name
+          return left.name - right.name;
```
RomanHotsiy:
```js
"test" - "test" === NaN
```
Maybe use `localeCompare` ?

RomanHotsiy:
LGTM 👍
Thanks!

PR #1102, opened 2019-11-13
[FEATURE] Add callback support

Jonathan-Bailey-Bose:
# Description
This PR utilizes the code from https://github.com/Redocly/redoc/pull/757 written by @Makashov to add callback support to ReDoc. It further adds to the code by adding the following:
- Adds `isCallback: boolean` field to the `OperationModel` class in order to distinguish a normal `Operation` from a `Callback`. This defaults to `false`.
- `Callback`s are defined with the following behavior as a type of `Operation`:
   - `Callback`s by default will use the `event name` for the view's label when rendered by ReDoc, as opposed to the `operationID`.
   - `Callback`s by default _do not inherit_ the specification's global `security` definition(s). They default to having no security, and can have security added on an individual basis.
   - `Callback`s by default _do inherit_ the specification's global `servers` definition(s). In many cases, this can be undesirable. Override the `servers` property on individual callbacks to remedy this if needed.
- Adds some tests to maintain code coverage.
- The rendering style of a callback is meant to mimic that of an operation within an operation response dropdown.

PR #757, opened 2018-12-26
Show callbacks

Makashov:
Show callbacks after **content item**. If callback has it's own callback it will be displayed recursively.

Trying to help with #411

PR #1224, opened 2020-04-02
feat: add callbacks support

okachynskyy:
This PR utilizes the code from #1102 written by @Jonathan-Bailey-Bose and  #757 written by @Makashov to add callback support to ReDoc.

- Updated layout for Callbacks middle section:
  - Callbacks by default will use the operation summary (or description, if summary does not exist) when rendered by ReDoc instead of callback name
  - Added method budge (`POST / PUT / etc.`) for callback summary
  - Added `Deprecated` label for deprecated callbacks (with strikeout callback summary)
image

- Created new component for visualizing Callback details
  - Show description in Callback details 
  - Move `Endpoint` component from the right panel to Callback details (added `compact` property for Endpoint component for displaying smaller HTTP method verb)
image

- By default, the callbacks do not use top-level Servers
image

- But servers can be defined
  - on the **path** level
image
  - or **operation** level
image

- Added one more dropdown for switching callback request examples (with truncating long names with an ellipsis). Created a "GenericChildrenSwitcher" component that tracks the selected item itself.
image

- Added new endpoint `/store/subscribe` and few callbacks in `openapi.yaml` demo file with different configs to illustrate all possible configurations for callbacks.

- Also, in the scope of this PR was updated styles for the label in the dropdown for consistency and added gray colors to the theme. 

----

### All together:
image


src/components/CallbackSamples/CallbackReqSamples.tsx
```diff
@@ -0,0 +1,35 @@
...
+  renderDropdown: (props: DropdownProps) => JSX.Element;
+}
+
+export class CallbackReqSamples extends React.Component<PayloadSamplesProps> {
+  render() {
+    const payloadSample = this.props.callback.codeSamples.find(sample =>
+      isPayloadSample(sample),
+    ) as XPayloadSample | undefined;
+
+    return (
```
RomanHotsiy, line 21: Return early here:

```js
if (!payloadSample) {
  return null;
}
```

src/components/CallbackSamples/CallbackReqSamples.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { OperationModel } from '../../services/models';
+import { XPayloadSample } from '../../services/models/Operation';
+import { isPayloadSample } from '../../services';
+
+export interface PayloadSamplesProps {
+  callback: OperationModel;
+  renderDropdown: (props: DropdownProps) => JSX.Element;
+}
+
+export class CallbackReqSamples extends React.Component<PayloadSamplesProps> {
```
RomanHotsiy, line 15: Rename to `CallbackPayloadSample`

src/components/CallbackSamples/CallbackSamples.tsx
```diff
@@ -0,0 +1,84 @@
...
+      .map(callback => callback.operations.map(operation => operation))
+      .reduce((a, b) => a.concat(b), []);
+
+    // Sums number of code samples per operation per callback
+    const numSamples = operations.reduce(
+      (sampleSum, operation) => sampleSum + operation.codeSamples.length,
+      0,
+    );
+
+    const hasSamples = numSamples > 0;
```
RomanHotsiy, line 44: Remove the reduce above and just use simple `Array.some` check

```suggestion
    const hasSamples = operations.some(operation => operation.codeSamples.length > 0);
```

RomanHotsiy: + return early here (`if (!hasSamples) { return null }`)

src/components/Callbacks/CallbackDetails.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return (
+      <CallbackDetailsWrap>
+        {hasDescription && (
+          <Description>
+            {description !== undefined && <Markdown source={description} />}
+            {externalDocs && <ExternalDocumentation externalDocs={externalDocs} />}
+          </Description>
+        )}
+        <Endpoint operation={this.props.operation} inverted={true} compact={true} />
+        {/* Do we need Extensions in callback details? */}
```
RomanHotsiy, line 35: yes, let it be

src/components/GenericDropdown/GenericDropdown.tsx
```diff
@@ -0,0 +1,70 @@
...
+  renderDropdown: (props: DropdownProps) => JSX.Element;
+  children: (activeItem: T) => JSX.Element;
+}
+
+export interface GenericDropdownState {
+  activeItemIdx: number;
+}
+
+@observer
+export class GenericDropdown<T> extends React.Component<
```
RomanHotsiy, line 20: I would rename this component to something like `GenericChildrenSwitcher` or maybe you have other ideas. `GenericDropdown` is not what this component is as it accept an option `renderDropdown` which is a bit confusing.

src/components/Responses/ResponsesList.tsx
```diff
@@ -3,6 +3,7 @@ import { ResponseModel } from '../../services/models';
 import styled from '../../styled-components';
 import { ResponseView } from './Response';
 
+// TODO: get text color and font size from the theme?
 const ResponsesHeader = styled.h3`
```
RomanHotsiy, line 7: Yes! Would be great! Change 18px to something in `em` and get color from theme.

src/components/GenericDropdown/GenericDropdown.tsx
okachynskyy: What about moving the logic of rendering dropdown/label into this GenerciDropdown component? And just passing Dropdown/Label as props?

RomanHotsiy:
🎉 

Thanks @Jonathan-Bailey-Bose and @Makashov for initial implementation!

Kudos to @okachynskyy for finishing it up 👏 🔥!

I will publish a new release at the end of the week.

RomanHotsiy:
Released in `2.0.0-rc.27`

PR #1229, opened 2020-04-13
Remove core-js/promise bundle into worker

kaushalmk90:
I had an issue with core-js/es/promise as it was unable to resolve. Looking at the code, I thought we could do away with this.  Just my thought..


RomanHotsiy:
What issues did you have? Could you provide more details? Could you open an issue? You can't remove this. It is for IE11 compatibility.

Issue #573, opened 2018-07-24
Selected value for discriminator field doesn't make sense if response is a subschema


lei-sumo:
For the `PetStore` example, if I add an endpoint to return a `Dog` by Id and reference `Pet`'s subschema `Dog`, I would expect the generated doc has `Dog` for the response schema. However, it is still showing `Pet` as the response type and default the selection to `Cat`. This is very confusing. As we already know its strong type, shouldn't it point to the correct value (and disable the dropbox ideally)?

Endpoint added:
```
'/dog/{dogId}':
    get:
      tags:
        - pet
      summary: Find dog by ID
      description: Returns a single dog
      operationId: getDogById
      parameters:
        - name: dogId
          in: path
          description: ID of dog to return
          required: true
          deprecated: true
          schema:
            type: integer
            format: int64
      responses:
        '200':
          description: successful operation
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Dog'
            application/xml:
              schema:
                $ref: '#/components/schemas/Dog'
        '400':
          description: Invalid ID supplied
        '404':
          description: Dog not found
      security:
        - api_key: []
```

Generated UI
![image](https://user-images.githubusercontent.com/15622224/43110155-39c78ade-8e9f-11e8-8511-81a9689a6362.png)

Expected UI
![image](https://user-images.githubusercontent.com/15622224/43110235-a8361c38-8e9f-11e8-9035-54ae1ee2329d.png)

I'm not too worried about it if only discriminator field is populated wrongly. But it actually renders other `Cat` specific fields and ignores `Dog` specific fields. This makes it unacceptable to release the documentation to our customers.

RomanGotsiy:
I will investigate this.
Thanks for the detailed issue report

lei-sumo:
Hi @RomanGotsiy, do you have any idea how to fix this?

RomanGotsiy:
Yes, I started working on this but don't have enough time to finish it.
It's a big change so have to be very careful.

lei-sumo:
Great. Thanks for the update.

crussell52:
Hi -- new user of ReDoc... loving it... but I immediately ran into this.  Is there an update on progress?

fabianvo:
I ran into the same issue and tried to fix it. Maybe my hack can help someone to come up with a better solution. Or at least start a discussion about improved discriminator support in ReDoc.

vadzim-marchanka:
[...]

PR #1231, opened 2020-04-15
#573 fix discriminator handling when referencing classes which only have a superclass, but no subclasses (and add description to ObjectSchema.tsx)

fabianvo:
I ran into an issue when rendering documentation for an API with polymorphism:

- the discriminator dropdown was always rendered, even if the type only had a superclass with a discriminator, but no subclasses
- also, any description was ignored for Schemas

I managed to fix the problem for my use case, but I am not sure if the solution is correct. I am still opening this pull request, maybe someone with more experience with redoc can use it to provide a better solution for the issue.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -50,11 +50,13 @@ $ npm run unit
 
 # run e2e tests
 $ npm run e2e
+# Make sure you have created bundle before running e2e test
+# E.g. run `npm run bundle` and wait for the finishing process.
 
 # open cypress UI to debug e2e test
 $ npm run cy:open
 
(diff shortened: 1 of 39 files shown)


RomanHotsiy:
Can you provide a minimal reproducible sample of your issue?

fabianvo:
We generate our API via Swagger Core. I stripped it down to show the core issue and mangled all names. Additionally, I included a small diagram to illustrate our inheritance hierarchy.

The discriminator dropdown is always displayed in the generated documentation, instead of only where it is needed (for `CDTO` and `MDTO`).

Hope this helps...

classes
[example.zip](https://github.com/Redocly/redoc/files/4604321/example.zip)

PR #1244, opened 2020-04-24
Add security headers to nginx config

Dalabad:
Increase the security of the nginx server and the served page, by adding the following security headers to the nginx config:
- X-Frame-Options (Disables click jacking by disallowing the page to be run in a frame/iframe)
- X-XSS-Protection (Enables cross site scripting filtering)
- X-Content-Type-Options (Disables MIME sniffing and forces browser to use the type given in Content-Type.)
- Content-Security-Policy (Controls resources the user agent is allowed to load for a given page.)
- Referrer-Policy (Governs which referrer information sent in the Referer header should be included with requests made.)

Additional headers that could be added optionally:
- Strict-Transport-Security (Enforce HTTPS over HTTP)

### DIFF
--- a/config/docker/nginx.conf
+++ b/config/docker/nginx.conf
@@ -17,6 +17,13 @@ http {
     server_name       localhost;
     index             index.html index.htm;
 
+    # Add security headers
+    add_header X-Frame-Options deny always;
+    add_header X-XSS-Protection "1; mode=block" always;
+    add_header X-Content-Type-Options nosniff always;
+    add_header Content-Security-Policy "default-src 'none'" always;
+    add_header Referrer-Policy strict-origin-when-cross-origin always;
+
     location / {
       alias            /usr/share/nginx/html/;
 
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
Thanks for the PR!

Have you tested if Redoc works with those headers enabled? I remember issue report related to `Content-Security-Policy` and web-workers so want to make sure you verified it works.

Dalabad:
I had to restructure the add_header commands, as nginx does not correctly inherit the headers from higher hierarchies.

Here is a running version of this branch with the added headers: [Redoc-PR-1244](https://docs.crashtest.cloud/)

So far I could not identify any problems with the added headers interfering in any way

RomanHotsiy:
Thanks!

I see this in the Chrome console: 
image

Also, could you please test it on localhost too? Usually, security headers behave funky when it comes to localhost

Dalabad:
Thanks for the console output.

You were right, the CSP seems to make problems if configured as the external sources like google fonts and swagger files would not load correctly. I've removed the CSP header for now.

I've updated the linked version from my previous commit. I've also tested it locally and did not get any more visible or console errors.

PR #1267, opened 2020-05-22
Updated x-servers link to point to valid location.

JoBrad:
The file `docs/redoc-vendor-extensions.md` contained a link to `https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.0.md#server-object`, which is invalid. I updated it to point to `https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject`


RomanHotsiy:
👍

PR #1287, opened 2020-06-01
feat: add JSON syntax highlighting

Quitra01:
Using JavaScript highlights for JSON looks confusing as keys and string values are the same color.

### DIFF
--- a/src/utils/highlight.ts
+++ b/src/utils/highlight.ts
@@ -8,6 +8,7 @@ import 'prismjs/components/prism-csharp.js';
 import 'prismjs/components/prism-go.js';
 import 'prismjs/components/prism-http.js';
 import 'prismjs/components/prism-java.js';
+import 'prismjs/components/prism-json';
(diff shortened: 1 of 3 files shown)


src/utils/highlight.ts
```diff
@@ -8,6 +8,7 @@ import 'prismjs/components/prism-csharp.js';
 import 'prismjs/components/prism-go.js';
 import 'prismjs/components/prism-http.js';
 import 'prismjs/components/prism-java.js';
+import 'prismjs/components/prism-json';
```
RomanHotsiy, line 11:
```suggestion
import 'prismjs/components/prism-json.js';
```

Quitra01:
Thank you, I hadn't realised that. I guess simply removing the mapping to `js` will be enough here since prismjs accepts `json`.

RomanHotsiy:
OK, I had to rework this after I tasted it locally.

Looks like below right now: 

image

PR #1300, opened 2020-06-12
Prevent body scrolling when user scrolls side menu

flaksp:
Added [`overscroll-behavior: contain`](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) to side menu when native scrollbars used. It will prevent body scrolling when user scrolls over menu using mouse wheel or trackpad.


RomanHotsiy:
Thanks!

PR #1302, opened 2020-06-16
fix: server list cut off (#980)

### LINKED ISSUE #980: Bottom of the docs gets cut-off
Even in the [Live Demo](http://redocly.github.io/redoc/#operation/logoutUser), if you scroll to the very bottom then expand the "Logout" request details:

image

It may have to do with simple requests (like logout) that do not have a lot of extra information returned by the server.

vishal-vp:
Fix #980


Oprysk:
fixed in #1803

Issue #1180, opened 2020-02-19
Side-menu : operation-type label width too short when display "PATCH" in Chrome.


mzone:
Side-menu : operation-type label width too short when display "PATCH" in Chrome.

Probably the cause by font-size. Chorme can't diplay font size less than 10px.

I want to change here.
```
# src/components/SideMenu/styled.elements.ts
# className: `operation-type ${props.type}`
-  width: 32px;
+  width: 6.5em;
```

| as-is | to-be |
|---|---|
| <img width="280" alt="74829928-6c265b00-5355-11ea-9945-13dd49b9a83a" src="https://user-images.githubusercontent.com/1255599/75016818-2b525180-54cf-11ea-9382-499e92830387.png"> | <img width="281" alt="74829974-806a5800-5355-11ea-93e2-4f166e25748c" src="https://user-images.githubusercontent.com/1255599/75016827-2ee5d880-54cf-11ea-8f72-98b46a90868c.png"> |

thanks..

RomanHotsiy:
It looks fine for me 
![image](https://user-images.githubusercontent.com/3975738/81507979-b93d1c00-9309-11ea-9acb-72949c0918fc.png)

Can you provide more details? What OS and monitor do you use?

mariusa:
![Screenshot from 2020-05-27 12-31-02](https://user-images.githubusercontent.com/1283842/83002742-12999080-a016-11ea-8931-72a2ed01edfd.png)

Linux, Chrome 82 with larger font size & minimum:
![Screenshot from 2020-05-27 12-33-09](https://user-images.githubusercontent.com/1283842/83002914-4674b600-a016-11ea-89ba-82bc34b18bf3.png)

mzone:
> It looks fine for me
> ![image](https://user-images.githubusercontent.com/3975738/81507979-b93d1c00-9309-11ea-9acb-72949c0918fc.png)
> 
> Can you provide more details? What OS and monitor do you use?

Sorry I haven't been able to reply.

OS: macOS Catalina 10.15.5
Browser: google Chrome 81.0.4044.138
MacBook Pro (15-inch, 2019)

still have that problem X<
Maybe, a problem with the font... I'm using japanese font 'font-family:HiraginoSans-W3' (typical japanese  font in mac)

thanks.

wopian:
Also facing this problem with the default settings (besides colours) when the header weight is modified.

```
--options.theme.typography.headings.fontWeight=700
```

`bold/700` font weight:

[...]

PR #1356, opened 2020-08-19
fix: scale sideMenu labels according to computed font size

wopian:
Closes https://github.com/Redocly/redoc/issues/1180.

See https://github.com/Redocly/redoc/issues/1180#issuecomment-676797731 for the `width` change.

See https://github.com/Redocly/redoc/issues/1180#issuecomment-634545588 for the `font-family` change.

---

Setting units in `px` and `(r)em` uses the CSS-defined `font-size` units (in this case `7px`) and not the computed `font-size`, which means the label size does not scale with the increased minimum font sizes users can set in their browser's settings. Minimum font size is a setting in pretty much every browser on Windows, Linux and MacOS and is likely set to `None` by default.

If a user sets the minimum font size above `7px` (for example `10px`) then the label width (a static `32px`) becomes too short for the text width.

The `ex` unit is supported by Chrome 1+, Edge 12+, Firefox 1+, IE 4+ and Safari 1+. 

|Description|`32px`|`9ex`
|-----------|------|-----
|`7px` font size and no minimum font size (`7px` computed) | image|image
|`7px` font size and a minimum font size of `10px` (`10px` computed) | image| image
|`7px` font size and a mimimum font size of `10px` (4K display, `7px` computed) | image | image

### DIFF
--- a/src/components/SideMenu/styled.elements.ts
+++ b/src/components/SideMenu/styled.elements.ts
@@ -7,7 +7,7 @@ import styled, { css, ResolvedThemeInterface } from '../../styled-components';
 export const OperationBadge = styled.span.attrs((props: { type: string }) => ({
   className: `operation-type ${props.type}`,
 }))<{ type: string }>`
-  width: 32px;
+  width: 9ex;
   display: inline-block;
   height: ${props => props.theme.typography.code.fontSize};
   line-height: ${props => props.theme.typography.code.fontSize};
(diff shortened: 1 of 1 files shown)


src/components/SideMenu/styled.elements.ts
```diff
@@ -16,7 +16,7 @@ export const OperationBadge = styled.span.attrs((props: { type: string }) => ({
   background-repeat: no-repeat;
   background-position: 6px 4px;
   font-size: 7px;
-  font-family: Verdana; // web-safe
+  font-family: Verdana, sans-serif; // web-safe
```
adamaltman, line 19: Yes, this is important!

adamaltman: This is what it looks like for me without it:

image

RomanHotsiy (approved):

RomanHotsiy:
This is awesome @wopian 👏 

Thanks for the detailed explanation!

wopian:
You're welcome 😀

PR #1378, opened 2020-09-08
feat: add optional BASE_PATH to Docker config

draoncc:
While working with redoc I came accross the lack of a "base path" option in the provided Docker image.

My use case is that I have redoc running behind a reverse proxy routing via path matching. E.g. the reverse proxy receives a request to `<host>/docs`, strips the request URL of `/docs` and forwards it to redoc. The way index.tpl.html is constructed, it tries to load favicon.png from `<host>/favicon.png`. The workaround is to send requests to `<host>/docs/` instead, where it'll be correctly loaded from `<host>/docs/favicon.png`, but this feels dirty.

Setting `BASE_PATH` to `/docs/` will prepend favicon.png, making it load from `<host>/docs/favicon.png` no matter the original request.

A possible alternative to this could be allowing users to specify BUNDLE_URL or similar, which defaults to `redoc.standalone.js`. That way, users could set BUNDLE_URL to `/docs/redoc.standalone.js` and PAGE_FAVICON to `/docs/favicon.png`. Another possible alternative could be to modify the nginx config.

I'd be happy to make adjustments.

### DIFF
--- a/config/docker/Dockerfile
+++ b/config/docker/Dockerfile
@@ -25,6 +25,7 @@ FROM nginx:alpine
 
 ENV PAGE_TITLE="ReDoc"
 ENV PAGE_FAVICON="favicon.png"
+ENV BASE_PATH=
 ENV SPEC_URL="http://petstore.swagger.io/v2/swagger.json"
 ENV PORT=80
 ENV REDOC_OPTIONS=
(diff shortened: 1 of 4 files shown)


tvestelind:
I have the same problem and this PR would solve it.

I think that the changes look good, but it's not up to me :sweat_smile:

RomanHotsiy (approved):

zalesky (approved):

AlexVarchuk:
We'll release the new version of docker next week.

Issue #915, opened 2019-05-15
Infer type from single allOf $ref


BenjaminPelletier:
Specifying the description for a property with a complex type is [not allowed](https://github.com/OAI/OpenAPI-Specification/issues/556); for example, this is invalid because the description will be wiped out by $ref replacement:
```
FooOperationURL:
    description: Endpoint that provides the ability to perform a foo operation
    type: string
```
```
preview_url:
    description: Endpoint at which to preview data
    $ref: '#/components/schemas/FooOperationURL
```
The suggested (messy & ugly, IMO) workaround for this is to specify the type in an allOf clause:
```
preview_url:
    description: Endpoint at which to preview data
    allOf:
        - $ref: '#/components/schemas/FooOperationURL
```
...but, ReDoc indicates that the type of `preview_url` in the latter case is merely `string` while in the former case it was much more usefully `string (FooOperationURL)`.

I suggest that ReDoc treat a singular allOf $ref as the implied type of the property and therefore print that type hint in the documentation.

Tecktron:
I agree, ReDoc is not handling the `allOf` properly correctly. It should expand every item inside within it including any references. This especially holds true in this case since this is the JSON Schema's suggested way of adding a description to a referenced object and is a very common use case for having reusable definitions.

RomanHotsiy:
Until this is fixed you can try using the following workaround:

```yaml
preview_url:
  allOf:
    - $ref: '#/components/schemas/FooOperationURL
    - description: Endpoint at which to preview data
```

mattiasrunge:
For me it seems the workaround with allOf does not work correctly without adding type and then not for array types.
Since we see this as a limitation of the specification, could it not be possible to support this case with an extension.

```yml
preview_url:
    x-description: Endpoint at which to preview data
    $ref: #/components/schemas/FooOperationURL
```

PR #1381, opened 2020-09-11
feat: add support for x-description to override description of resolved

mattiasrunge:
To solve issues like #915 this could be a simple solution.
It overrides the resolved $ref's description with what is set in x-description.
```yml
preview_url:
    x-description: Endpoint at which to preview data
    $ref: #/components/schemas/FooOperationURL
```

### DIFF
--- a/src/services/OpenAPIParser.ts
+++ b/src/services/OpenAPIParser.ts
@@ -164,6 +164,9 @@ export class OpenAPIParser {
         this.exitRef(resolved);
         return res;
       }
+      if (obj['x-description']) {
+        return Object.assign({}, resolved, { description: obj['x-description'] });
+      }
       return resolved;
     }
     return obj;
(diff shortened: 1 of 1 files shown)


adamaltman:
In the example above you show a sibling ref. From the spec, about the reference object:

> This object cannot be extended with additional properties and any properties added SHALL be ignored.

mattiasrunge:
Yes, but it does say: "any properties added SHALL be ignored" which in my view means it would not break anything. 
It also says: 
> The Reference Object is defined by JSON Reference and follows the same structure, behavior and rules.

The JSON Reference (https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) spec only states that additional members should be ignored, not that it is forbidden to extend the object.
> Any members other than "$ref" in a JSON Reference object SHALL be ignored.

So I agree that it is not ideal with an x-description, but the OpenAPI spec says it follows the JSON Reference spec which does not forbid it, that should take precedence.

hleb-albau:
it is 3.1 openapi spec feature! But it will take some time, to adopt tool set(parsers and etc)

RomanHotsiy:
We now have basic support for OpenAPI 3.1 with this feature supported.

PR #1394, opened 2020-09-28
Davew/bdevexp 2652 redoc json coloring

dave-wheeler-OT:
setting a couple of "on dark" colors for json/code displays to meet accessibility contrast guidlines


dave-wheeler-OT:
i did not mean for the master to be the target.  I intended the 1200 branch.

PR #1406, opened 2020-10-09
Added support for showExtensions and $ref for Responses

shelbys:
If you appreciate this contribution, please add a `hacktoberfest-accepted` Label to the PR, so it'll be tracked

### DIFF
--- a/src/services/__tests__/models/Response.test.ts
+++ b/src/services/__tests__/models/Response.test.ts
@@ -39,5 +39,12 @@ describe('Models', () => {
       const resp = new ResponseModel({ ...props, code: 'default', defaultAsError: true });
       expect(resp.type).toEqual('error');
     });
+
+    test('should be error if showExtensions is true', () => {
+      const options = new RedocNormalizedOptions({ showExtensions: true });
+      const resp = new ResponseModel(parser, 'default', true, { 'x-example': {a: 1} } as any, options);
+      expect(Object.keys(resp.extensions).length).toEqual(1);
+      expect(resp.extensions['x-example']).toEqual({a: 1});
+    });
   });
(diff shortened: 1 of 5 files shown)


src/services/__tests__/models/Response.test.ts
```diff
@@ -31,5 +31,12 @@ describe('Models', () => {
       const resp = new ResponseModel(parser, 'default', true, {}, opts);
       expect(resp.type).toEqual('error');
     });
+
+    test('should be error if showExtensions is true', () => {
```
RomanHotsiy, line 35: should be error?

RomanHotsiy (changes requested):
It looks good, the only question is if the test message is corret?

src/services/__tests__/models/Response.test.ts
gileswells, line 35: Maybe this?

```suggestion
    test('ensure extensions are shown if showExtensions is true', () => {
```

RomanHotsiy (approved):

src/services/__tests__/models/Response.test.ts
shelbys: Oops!  I missed updating the first half.  ;)

shelbys:
Thanks much, including for the tagging

shelbys:
@RomanHotsiy - Please merge this when you get a chance to unblock #1434

RomanHotsiy:
Hey @shelbys!

I am very sorry for the 1-year delay! 🙈
We've been busy with other projects and I missed this PR (and many more others 😢 ).

We now have more resources and time to work on Redoc so this PR will be merged today.

PR #1423, opened 2020-10-18
Add field constraint indicator for uniqueItems

### LINKED ISSUE #1353: uniqueItems: true is not exposed in generated docs
For arrays, setting uniqueItems to true is not reflected in the documentation, but should generally be considered important for a reader.

~~~~~~~ yaml
    type: array
    items:
      type: string
      enum:
        - "one"
        - "two"
        - "three"
    minItems: 1
    uniqueItems: true
~~~~~~~

Screenshot from 2020-08-18 12-27-05

Leaving this to the description leads to inconsistent and insufficient documentation.

miqh:
Hello,

Fixes #1353.

Schemas which define `uniqueItems` should now generate a field constraint indicator.

For the example provided in the linked issue, the visual output looks like the following:

Screenshot from 2020-10-18 09-42-51

---
I also noticed `ArraySchema` tries to indicate array-type constraints. Specifically, it only covers the `minItems` and `maxItems` constraints.

https://github.com/Redocly/redoc/blob/1fdfd1353dfc07aca5f5d55e5e2195648bde6e2b/src/components/Schema/ArraySchema.tsx#L21

I haven't dug too deeply into this, but with a simple schema such as:

```yaml
type: array
  items:
    type: number
  minItems: 1
```

I'd expect to see something like the following (based on the code), but I don't:

Screenshot from 2020-10-18 10-13-56

Anyway, I'm happy to follow-up and do something about the latter if it's indeed unexpected behaviour.

### DIFF
--- a/src/components/Fields/FieldContstraints.tsx
+++ b/src/components/Fields/FieldContstraints.tsx
@@ -14,7 +14,7 @@ export class ConstraintsView extends React.PureComponent<ConstraintsViewProps> {
       <span>
         {' '}
         {this.props.constraints.map(constraint => (
-          <ConstraintItem key={constraint}>{constraint}</ConstraintItem>
+          <ConstraintItem key={constraint}> {constraint} </ConstraintItem>
         ))}
       </span>
     );
--- a/src/components/Fields/FieldContstraints.tsx
+++ b/src/components/Fields/FieldContstraints.tsx
@@ -14,7 +14,7 @@ export class ConstraintsView extends React.PureComponent<ConstraintsViewProps> {
       <span>
         {' '}
         {this.props.constraints.map(constraint => (
-          <ConstraintItem key={constraint}> {constraint} </ConstraintItem>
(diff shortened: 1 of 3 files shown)


src/components/Fields/FieldContstraints.tsx
```diff
@@ -14,7 +14,7 @@ export class ConstraintsView extends React.PureComponent<ConstraintsViewProps> {
       <span>
         {' '}
         {this.props.constraints.map(constraint => (
-          <ConstraintItem key={constraint}> {constraint} </ConstraintItem>
+          <ConstraintItem key={constraint}>{constraint}</ConstraintItem>
```
RomanHotsiy, line 17:
```suggestion
          <ConstraintItem key={constraint}> {constraint} </ConstraintItem>
```

RomanHotsiy (commented):
LGTM

RomanHotsiy (approved):

PR #1424, opened 2020-10-18
fix #1286: Expand/Collapse all buttons disappears for flat structures

### LINKED ISSUE #1286: Expand all and Collapse all should be disabled for flat structures
What if a case that my example or schema model doesn't contains a array and my object is pure single dimensional.

For example:
image

we should disable expand all,collapse all options for flat objects. **Expand All**, **Collapse All** doesn't work in some cases and above is best example. user may feel that something wrong and not working.

AvroraPolnareff:
Hi, I've done work on issue #1286. I think toggle only buttons visibility is easier way to implement this. Let me know if you prefer something else.


src/components/JsonViewer/JsonViewer.tsx
```diff
@@ -26,27 +26,39 @@ class Json extends React.PureComponent<JsonProps> {
...
-            dangerouslySetInnerHTML={{
-              __html: jsonToHTML(this.props.data, options.jsonSampleExpandLevel),
-            }}
-          />
-        )}
-      </OptionsContext.Consumer>
-    </JsonViewerWrap>
-  );
+  renderInner = ({ renderCopyButton }) => {
+    const showFoldingButtons = Object.values(this.props.data).some(
```
RomanHotsiy, line 30:
```suggestion
    const showFoldingButtons = this.props.data && Object.values(this.props.data).some(
```

src/components/JsonViewer/JsonViewer.tsx
```diff
@@ -26,27 +26,39 @@ class Json extends React.PureComponent<JsonProps> {
...
+      (value) => typeof value === 'object',
```
RomanHotsiy, line 31:
```suggestion
      (value) => typeof value === 'object' && value !== null,
```

src/components/__tests__/JsonViewer.tsx
```diff
@@ -42,5 +42,16 @@ describe('Components', () => {
 
       expect(ClipboardService.copySelected as jest.Mock).toHaveBeenCalled();
     });
+
+    test('Expand/Collapse buttons disappears for flat structures', () => {
+      const flatData = { a: 1, b: '2' };
```
RomanHotsiy, line 47:
```suggestion
      const flatData = { a: 1, b: '2', c: null };
```

RomanHotsiy (changes requested):
Suggested a few corrections to properly handle `null` values

src/components/JsonViewer/JsonViewer.tsx
AvroraPolnareff: Thank for suggestion! I tested it, all works fine.

AvroraPolnareff (commented):
All works fine, I've added suggestions to pull request.

zalesky:
Screenshot 2022-04-22 at 11 48 08

LGTM

@AvroraPolnareff, Thanks for your input 👍

zalesky (approved):

zalesky:
@RomanHotsiy  Would you kindly unblock this?

RomanHotsiy (approved):

Issue #1417, opened 2020-10-14
redoc taking 5 or more minutes to open my documentation


danilocastelhano1:
Hi everyone, i have an huge API with many endpoints, and these endpoints have a lot of fields, and the redoc is taking 5 min or more to open my documentation, is there anything i can do to make it faster? thank you.

stueynz:
We had similar problems; 47 data objects x 7 endpoints each (POST + GET, PUT & PATCH (single & multiple) ... and that's  just covering the CRUD operations; we'll have a bunch more for  more RPC based endpoints.   

In the end we chopped the spec up into seven sections, used redoc-cli to produce seven matching static HTML files, and then glued them back together with links across the boundaries.

https://sifnzmodel.azurewebsites.net/OpenAPI-v3.2-rc2-2020-10-07/TeRito-OpenAPI-organisation.html

danilocastelhano1:
When i try to run redoc-cli bundle on my spec, i got this error: FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

lornajane:
Since this issue has been open a while, it's difficult to reproduce. Please let us know if you're still seeing performance problems after upgrading to the newest Redoc release.

Issue #1397, opened 2020-09-30
redocly.github.io/redoc broken -> Error downloading https://cors.apis.guru/...


ajmas:
When trying the demo with the following swagger files:

  - https://api.apis.guru/v2/specs/googleapis.com/calendar/v3/swagger.yaml
  - https://api.apis.guru/v2/specs/slack.com/1.2.0/swagger.yaml

The following error is generated:

```
Error downloading https://cors.apis.guru/https://api.apis.guru/v2/specs/googleapis.com/calendar/v3/swagger.yaml HTTP ERROR 404
Stack trace
Error: Error downloading https://cors.apis.guru/https://api.apis.guru/v2/specs/googleapis.com/calendar/v3/swagger.yaml 
HTTP ERROR 404
    at https://redocly.github.io/redoc/redoc-demo.bundle.js:48:1501
    at https://redocly.github.io/redoc/redoc-demo.bundle.js:21:86584

Error: HTTP ERROR 404
    at https://redocly.github.io/redoc/redoc-demo.bundle.js:48:1501
    at https://redocly.github.io/redoc/redoc-demo.bundle.js:21:86162

ReDoc Version: 2.0.0-rc.40
Commit: 17b9873
```

Additionally, when the error happens clicking 'try it' with a new URL results in no apparent action, even the URL in the address bar is updated. No error in the console either.

Note, it works with:

  - https://api.apis.guru/v2/specs/zoom.us/2.0.0/swagger.yaml
  - https://docs.graphhopper.com/openapi.json

Testing with Chome 86.0.4240.55 on MacOS X 10.15.6

ajmas:
Further investigation reveals that the following pages are returning a 404:

- https://api.apis.guru/v2/specs/googleapis.com/calendar/v3/swagger.yaml
- https://api.apis.guru/v2/specs/slack.com/1.2.0/swagger.yaml

Error handling should be updated to test for a 404.

adamaltman:
Thanks for the report. We have a fix in the works.

adamaltman:
@pshenok did you open a PR?

RomanHotsiy:
@adamaltman yes, but the demo was not redeployed for some reason. I just redeployed it manually.

PR #1436, opened 2020-10-27
fix: lazy fields to prevent crash on complex nested oneOf's

### LINKED ISSUE #696: Schema causing ReDoc to run out of memory
We have a rather large schema that is causing Redoc to fail to render as the browser runs out of memory.

I have tried utilising redoc-cli to bundle but this will run seemingly forever and never complete.

The schema can be found here: https://gist.github.com/AlexRichter/001f64d9c5ac17a8a0bfdb1a771eef71

I have tried creating a minimal use case but can't as I believe the issue is due to the complexity of the schema.

You can fix it by removing some of the `oneOf` and `discriminator` in `CheckResource`. If you only have 3 it will render quickly, up to 5 and it takes ~18 seconds. All of them and it will never complete.

Your help is much appreciated!

RomanHotsiy:
Fixes #696


jfoliveira:
Hey @knidarkness @okachynskyy as this seems to fix #696 #1176 #1397 and #1417 could you please review?
Just a heads up as you guys are the only ones marked as reviewers.

BobbyJohansen:
Any Movement on this? We have some pretty large specs that we need to churn through

glen-84:
I tried this with our spec which is ~28,000 lines, and:

1. `serve` uses up all available memory.
2. `bundle` hits a `RangeError: Invalid string length` at objectToHTML (.../redoc.lib.js:11543:19).

😞

prabhutw:
@RomanHotsiy When can we expect this PR to be merged? We are also facing performance issue because of oneOf
This PR will fix #1814 as well

RomanHotsiy:
This PR is not ready to merge, unfortunately. It causes many side effects across the app state so we have to come up with a better solution.

PR #1443, opened 2020-10-30
Add the title tag to the Array () definition within a schema block

cread-creditsense:
When defining multi-dimensional arrays within a schema, the "Array ()" blocks do not currently show any attached title information about the array levels. This PR adds the contents of a title element just underneath each "Array ()" level to assist readability. This provides a layout like the following:

image


adamaltman:
Nice idea.

src/components/Schema/ArraySchema.tsx
```diff
@@ -23,6 +24,7 @@ export class ArraySchema extends React.PureComponent<SchemaProps> {
     return (
       <div>
         <ArrayOpenningLabel> Array ({minMaxItems})</ArrayOpenningLabel>
+        <TypeTitle>{itemsSchema.schema.title}</TypeTitle>
```
adamaltman, line 27: what if there is no title?

cread-creditsense: Good call, set the block to only render if the schema title exists.

RomanHotsiy:
This is similar to https://github.com/Redocly/redoc/pull/1497

We will definitely consider this feature for this but a little bit later. Before, we want to clean up other PRs and release 2.0.

lornajane:
Closing in favour of #1497 , but thank you very much for being part of the project (and sorry for the very, very slow response)

Issue #1169, opened 2020-02-06
Add a lock file for redoc-cli


baohx2000:
Just a FYI when people start trying to submit bugs about this, mobx-react shipped a funky build which is causing broken-ness (at least when using redoc-cli bundle)

https://github.com/mobxjs/mobx-react/issues/831

Currently breaks with an error like:
`Cannot find module '..../.npm/_npx/1868/lib/node_modules/redoc-cli/node_modules/mobx-react/dist/mobxreact.js` (added for searchability)

Can redoc-cli ship with a lock file so this kind of thing doesn't happen?

enzobrands:
The original issue mobxjs/mobx-react#831 was fixed in a new release 6.1.7. The redoc-cli is working again. 
+1 for the lock file though.

baohx2000:
Changed issue name so this can be a request for a lock file :)

RomanHotsiy:
The lock file is not used when you install some dependency from npm, only when your run `npm install` in the folder.

So, I can't do this. The only idea is to pinpoint a specific version of `mobx-react` and other deps. What do you think?

baohx2000:
Redoc-cli isn't a dependency though, it's an application itself, which needs to be consistent in its operation. When you tag a release on npm does npm lock the dependency versions? Doesn't seem like it, which is why the tool broke when someone broke the dependency. Maybe I'm missing something?

RomanHotsiy:
> When you tag a release on npm does npm lock the dependency versions

No. When you install a redoc-cli using `npm install` or `npx` npm uses only `package.json` to resolve dependencies. It doesn't look at npm-shrinkwrap.

Shrinkwrap is used if only you clone this repo, go to the `cli` folder and run npm install there. We have `package-lock.json` in `cli` folder file there for this purpose.

RomanHotsiy:
Oh, sorry. I was wrong. I just investigated it and looks that shrinkwrap is actually used for locking when installing from npm.
I'll merge the PR by @calebboyd.

PR #1446, opened 2020-10-30
fix: use shrinkwrap for cli package

calebboyd:
#1169


baohx2000:
wow...is this really all that's required to fix that?  As long as npm / npx uses that file automatically for locking, I'll be happy.

RomanHotsiy (approved):

PR #1491, opened 2021-01-04
H2: Set color form theme.colors.primary.main

ajeetshah:
H2: Set color form theme.colors.primary.main so that it can be customized.

### DIFF
--- a/src/common-elements/headers.ts
+++ b/src/common-elements/headers.ts
@@ -22,7 +22,7 @@ export const H1 = styled.h1`
 
 export const H2 = styled.h2`
   ${headerCommonMixin(2)};
-  color: black;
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
It would make sense to add an extra theme toggle for h1 and h2 tags instead of changing the color of h2. Right now h2 is black but will get primary color when this PR is merged which is may be not the desired behavior for every Redoc user.

What do you think?

RomanHotsiy (dismissed):
Support a separate theme toggle.

west1cles:
An extra theme toggle for H1, H2 and H3 would be 👍

AlexVarchuk (approved):

PR #1500, opened 2021-01-17
chore: use openapi-core to bundle definition instead of json-schema-ref-parser

andriyl:
Use openapi-core to bundle definition instead of json-schema-ref-parser


teerawat111Thai (changes requested):
> 

> 

>

src/utils/loadAndBundleSpec.ts
```diff
@@ -1,19 +1,30 @@
-import * as JsonSchemaRefParser from 'json-schema-ref-parser';
 /* tslint:disable-next-line:no-implicit-dependencies */
 import { convertObj } from 'swagger2openapi';
 import { OpenAPISpec } from '../types';
+import { Source, Document, bundle, Config } from '@redocly/openapi-core';
```
RomanHotsiy, line 4:
```suggestion
import { Source, Document, bundle, Config } from '@redocly/openapi-core';
/* tslint:disable-next-line:no-implicit-dependencies */
import { convertObj } from 'swagger2openapi';
import { OpenAPISpec } from '../types';
```

RomanHotsiy (approved):

Issue #1463, opened 2020-11-13
Example drop-down cuts off text


andrewdialpad:
Minor visual issue, but the example drop-down under Response samples breaks visually with some text.

Example screenshot:
![image](https://user-images.githubusercontent.com/25596730/99096225-e6090f00-258a-11eb-805d-c26809cc1e4f.png)

PR #1508, opened 2021-01-26
Quick fix for #1463: Example drop down cuts off text

vkelembet:
Hi,
this is a quick fix for #1463 styling issue by setting value container line height to `normal`.

It looks like the dropdown component may have some search capabilities, which may be impacted by this change. Please point me to the page where I can spot it in the wild to test.

Also running `npm run prettier` (as per contribution guidelines), introduced a huge 72 files change, so I'm leaving mine without it for brevity.


anastasiia-developer:
@vkelembet Hi!)) Sorry for the huge delay. Thank you for your contribution to redoc.
We remove dropdown-aria and use native select  #1954

PR #1516, opened 2021-02-03
Add options to sort tags,operations and groups alphabetically

rinpatch:
Note: I did test that it behaves correctly when the if clauses are manually set to true, however wasn't able to test the options work since there doesn't seem any easy way to change options in the devserver or build `redoc-cli` using local redoc (at least as far as I could tell, some contribution documentation would be nice)


alexandrbarabolia:
@rinpatch add, please, options sort by type, example post, get, put, patch

rinpatch:
@alexandrbarabolia This is not something I personally have a need for, so no. Please add it yourself, pay someone to do it or open an issue in the repo and wait for someone to maybe pick it up.

RomanHotsiy:
We have implemented two options a while ago: `sortOperationsAlphabetically` and `sortTagsAlphabetically`.

Sorry for the delay.

PR #1533, opened 2021-02-14
disableDefaultSample option

lysenko86:


src/services/models/MediaType.ts
```diff
@@ -74,12 +74,22 @@ export class MediaTypeModel {
         );
       }
     } else if (this.schema) {
+      const sampledData = Sampler.sample(info.schema, samplerOptions, parser.spec);
+
+      if (disableDefaultSample && info.schema) {
```
RomanHotsiy, line 79: Instead of deleting the default sample, would it be easier to skip generating it?

Could you change it?

RomanHotsiy (changes requested):

Marshevskyy:
looks good, please check comments, also please update readme

PR #1536, opened 2021-02-16
demo: add support for getting spec from clipboard

### LINKED ISSUE #1251: Add a file selector as well as URL input for the demo
To make it easier to develop an openapi file locally, would it be possible to allow the user to upload a file, as an alternative to entering a URL?

davidjgoss:
image

Adds a button that, when clicked, will try to read the clipboard and import the spec from that.

- Uses the modern [async clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) (asks the user's permission first)
- Won't be shown in older browsers where it's not supported
- Uses the YAML parser so importing either YAML or JSON will work

Maybe closes #1251 - addresses the use case indirectly.


adamaltman:
Interesting idea. I think we would need to update that hint text to say they can paste it in too.

davidjgoss:
@adamaltman good point. Maybe something like:

> Enter a spec URL to try, or hit the paste icon to pull from clipboard

(Struggling to find more concise wording.)

Could also look at giving the button a short label like "Import from Clipboard"?

davidjgoss:
@adamaltman take 2:

image

Let me know what you think.

davidjgoss:
Any feedback on the updates here?

davidjgoss:
Given https://github.com/Redocly/redoc/pull/1859, this might be redundant now?

Issue #1563, opened 2021-03-31
Pin styled-components version until `ReferenceError: window is not defined` fixed


pierot:
See: https://github.com/styled-components/styled-components/issues/3444
If it's possible to pin the version number of styled-components in the mean time.

RomanHotsiy:
@pierot I'm not sure why you see the error because redoc-cli uses a npm-shrinkwrap file which should prevent this issue.

How do you run the redoc-cli?

I just tried `npx redoc-cli@0.10.4 bundle https://redocly.github.io/redoc/openapi.yaml -o petstore.html` and it worked on my machine.

Maybe I have some cache though.

flixx:
We've got the same error since today in our Pipelines...

```
------                                                                          
 > [apidoc 6/6] RUN redoc-cli bundle     --options.pathInMiddlePanel     -o /data/build/index.html     -t custom_template.hbs     /data/openapi/published-specs.yaml:                                                                           
#34 1.571 Prerendering docs                                                     
#34 3.167 ReferenceError: window is not defined
#34 3.167     at k (/usr/local/share/.config/yarn/global/node_modules/styled-components/dist/styled-components.cjs.js:1:7161)
#34 3.167     at e._emitSheetCSS (/usr/local/share/.config/yarn/global/node_modules/styled-components/dist/styled-components.cjs.js:1:22249)
#34 3.167     at e.getStyleTags (/usr/local/share/.config/yarn/global/node_modules/styled-components/dist/styled-components.cjs.js:1:22428)
#34 3.167     at /usr/local/share/.config/yarn/global/node_modules/redoc-cli/index.js:216:25
#34 3.167     at Generator.next (<anonymous>)
#34 3.167     at fulfilled (/usr/local/share/.config/yarn/global/node_modules/redoc-cli/index.js:6:58)
------
```

RomanHotsiy:
How is redoc-cli installed in the pipeline? using npm or maybe yarn? Also what version do you run in pipelines?

Sorry for maybe dumb questions, I'm just trying to understand how this happens provided we have npm-shrinkwrap in place.

Thanks.

flixx:
@RomanHotsiy No dumb question. Here you go:

```
FROM alpine:latest as apidoc
RUN apk --no-cache add git yarn && \
    yarn global add redoc-cli
COPY ./apidocs/openapi/ /data/openapi/
COPY ./apidocs/custom_template.hbs .
RUN mkdir -p /data/build

[...]

PR #1564, opened 2021-03-31
Set styled-componenets version to 5.2.1 until `ReferenceError: window is not defined` is fixed

Sokre95:
Issue page: #1563 

The `redoc-cli bundle` does not work with the 5.2.2 version of styled-components, while the `serve` works normally which is expected

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -20,7 +20,7 @@
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "redoc": "2.0.0-rc.50",
-    "styled-components": "^5.1.1",
+    "styled-components": "5.2.1",
     "tslib": "^2.0.0",
     "yargs": "^15.4.1"
   },


RomanHotsiy:
But the CLI has a shrinkwrap file with pinned version: https://github.com/Redocly/redoc/blob/master/cli/npm-shrinkwrap.json#L1782

Does it crash for you?

How do you run it?

Sokre95:
Hm, yes, it crashes to me too, obviously the shrinkwrap.json has no effect 😕
This is how I install & run cli
```
yarn add global redoc-cli
yarn run redoc-cli bundle
```

RomanHotsiy:
Could you please also try using npx or npm?

RomanHotsiy:
Fixed in styled components: https://github.com/styled-components/styled-components/pull/3446#issuecomment-811102602

PR #1566, opened 2021-04-03
Add nonce support

Okhoshi:
Redoc is broken if the page where it loads has CSP, in particular `style-src` directives. If inline styles are forbidden by the CSPs, the browser will refuse to load them, and Styled-Components won't work at all.

Styled-components already support CSPs by reading a nonce from `__webpack_nonce__` to apply it on every injected style tag. However, there's no way to configure this nonce with Redoc.

This is what I added in this PR. The nonce can either be filled in the HTML with a `nonce` attribute (similarly to other HTML elements like script or styles) or provided in the options object.


README.md
```diff
@@ -256,6 +256,7 @@ You can use all of the following options with standalone version on <redoc> tag
 * `payloadSampleIdx` - if set, payload sample will be inserted at this index or last. Indexes start from 0.
 * `theme` - ReDoc theme. For details check [theme docs](#redoc-theme-object).
 * `untrustedSpec` - if set, the spec is considered untrusted and all HTML/markdown is sanitized to prevent XSS. **Disabled by default** for performance reasons. **Enable this option if you work with untrusted user data!**
+* `nonce` - if set, the provided value will be injected in every injected HTML element in the `nonce` attribute. Useful with of CSP.
```
RomanHotsiy, line 259:
```suggestion
* `nonce` - if set, the provided value will be injected in every injected HTML element in the `nonce` attribute. Useful with of CSP, see https://webpack.js.org/guides/csp/
```

src/components/RedocStandalone.tsx
```diff
@@ -43,6 +45,10 @@ export class RedocStandalone extends React.PureComponent<RedocStandaloneProps> {
 
     const normalizedOpts = new RedocNormalizedOptions(options);
 
+    if (normalizedOpts.nonce !== undefined) {
+      __webpack_nonce__ = normalizedOpts.nonce;
```
RomanHotsiy, line 49: It may crash with `ReferenceError` in non-webpack env, e.g. when used by redoc-cli. We should add a check if the `__webpack_nonce__` is defined.

src/components/RedocStandalone.tsx
```diff
@@ -14,6 +14,8 @@ export interface RedocStandaloneProps {
   onLoaded?: (e?: Error) => any;
 }
 
+declare let __webpack_nonce__: string;
```
RomanHotsiy, line 17: Could you please move it to the custom.d.ts file?

RomanHotsiy (changes requested):
Looks great!

Could you please resolve a few minor comments? Thanks!

src/components/RedocStandalone.tsx
Okhoshi: Actually, this is already defined in the index.d.ts of the Webpack package.

src/components/RedocStandalone.tsx
Okhoshi: I had to use `try/catch` because `__webpack_nonce__` is initially set to `undefined` making disambiguation between variable being declared but not set and not being declared at all very hard.

If you have a better way, I'm open to suggestion.

Okhoshi:
@RomanHotsiy Thanks for the review ! I made the requested changes.

Okhoshi:
ping @RomanHotsiy

README.md
```diff
@@ -256,6 +256,7 @@ You can use all of the following options with standalone version on <redoc> tag
 * `payloadSampleIdx` - if set, payload sample will be inserted at this index or last. Indexes start from 0.
 * `theme` - ReDoc theme. For details check [theme docs](#redoc-theme-object).
 * `untrustedSpec` - if set, the spec is considered untrusted and all HTML/markdown is sanitized to prevent XSS. **Disabled by default** for performance reasons. **Enable this option if you work with untrusted user data!**
+* `nonce` - if set, the provided value will be injected in every injected HTML element in the `nonce` attribute. Useful with of CSP, see https://webpack.js.org/guides/csp/.
```
adamaltman, line 259:
```suggestion
* `nonce` - if set, the provided value will be injected in every injected HTML element in the `nonce` attribute. Useful with a CSP, see https://webpack.js.org/guides/csp/.
```

Okhoshi: Thanks, I rewrote the end of the sentence 👍

AlexVarchuk (approved):

PR #1578, opened 2021-04-12
Added missing semicolon

olivierbeaulieu:
There was a missing semicolon in the CSS. As is, it works without any issue, but if any minification happens, it can result in this:

image


RomanHotsiy (approved):

RomanHotsiy:
Thanks!

Issue #1048, opened 2019-09-30
SVG logo not displayed in Firefox


Trane9991:
Hello.
We are having issue when SVG logo is missing when docs opened in Firefox (in Chrome and Safari everything is fine), see [api link](https://docs.getsocial.im/api/).

The current version is v2.0.0-rc.14, however, I tried locally on rc15 and issue is still present.

Thanks!

RomanHotsiy:
Looks like this is issue with your SVG. Try setting width and height in you svg: https://stackoverflow.com/questions/21738612/svg-img-has-zero-dimensions-when-floated-in-firefox

PR #1597, opened 2021-04-30
fix: logo svg doesn't display or not a 100%

imagoiq:
Two issues with logo in SVG when using a link on it as the link is using `display: inline-block`:

- An SVG without height and width attributes display at 0 x 0 size
- An SVG with height and width attributes will only display at the given size and not at `width: 100%` as intended.

Fix this wrongly closed issue: https://github.com/Redocly/redoc/issues/1048

### DIFF
--- a/src/components/ApiLogo/styled.elements.tsx
+++ b/src/components/ApiLogo/styled.elements.tsx
@@ -14,7 +14,8 @@ export const LogoWrap = styled.div`
 `;
 
 const Link = styled.a`
-  display: inline-block;
+  display: flex;
+  justify-content: center;
 `;
 
 // eslint-disable-next-line react/display-name


Oprysk (approved):

zalesky:
https://user-images.githubusercontent.com/16255408/158361386-8278e0f5-b5e7-4d38-a9ac-4c5038963e02.mov

zalesky:
> @zalesky is it Firefox or Chrome on your video?

same behavior for both browsers

AlexVarchuk:
@imagoiq Hi! Thanks for your contribution.

But seems problem with your SVG file. Please check this [link](https://stackoverflow.com/questions/21738612/svg-img-has-zero-dimensions-when-floated-in-firefox). 
In this regard, we cannot merge it now 🙈 .

PR #1598, opened 2021-04-30
feat: redoc-cli add host option

imagoiq:
redoc-cli: add the ability to define `host` with option `-h` or `--host`

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -61,6 +61,12 @@ YargsParser.command(
       type: 'boolean',
     });
 
+    yargs.option('h', {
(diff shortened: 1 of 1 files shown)


colinmollenhour:
It would be better if the default host was "localhost" instead of "127.0.0.1" since on Windows if you run the cli via WSL the IP address link is broken but it works fine using the hostname.

Oprysk (approved):

Oprysk:
Thanks for your contribution! We are going to release it in the nearest feature.

PR #1640, opened 2021-06-09
feat: merge refs oas3.1

RomanHotsiy:


AlexVarchuk (approved):

src/services/OpenAPIParser.ts
```diff
@@ -58,8 +58,7 @@ export class OpenAPIParser {
     this.preprocess(spec);
 
     this.spec = spec;
-
-    this.mergeRefs = new Set();
+    this.allowMergeRefs = spec.openapi.startsWith('3.1');
```
ziadsawalha, line 61: Won't this break when `3.2` comes out? Should it be `>=`?

RomanHotsiy: I don' think so. OpenAPI doesn't follow semver so technically they can change the behaviour in 3.2. We'll update this when 3.2 is out.

Issue #1631, opened 2021-05-31
Handle emtpy object in security array


cmeeren:
According to https://github.com/OAI/OpenAPI-Specification/issues/1684, the following is valid:

```yaml
security:
  - Session: []
  - {}
```

(This is needed e.g. to signal that an operation is available both with a specific auth scheme and without authentication, e.g. for limited guest access.)

ReDoc currently renders it as this:

![image](https://user-images.githubusercontent.com/7766733/120190706-430f9e00-c219-11eb-96cc-c9eae386e993.png)

The rendering should be improved. Not sure what the best rendering would be, but anything that clearly says or indicates "no authentication" would already be better than nothing.

Sorry I can't contribute myself; I'm not a JS/front-end dev.

cmeeren:
Has this been released? I still see the old behavior.

cmeeren:
For future readers, seems like this works now:

![image](https://user-images.githubusercontent.com/7766733/143872412-61b5cbd3-74ed-45bb-b032-efb355d563e6.png)

PR #1678, opened 2021-07-07
fix: Handle emtpy object in security array

valdirmendesdev:
Fixes [#1631](https://github.com/Redocly/redoc/issues/1631)

### DIFF
--- a/src/components/SecurityRequirement/SecurityRequirement.tsx
+++ b/src/components/SecurityRequirement/SecurityRequirement.tsx
@@ -8,8 +8,8 @@ import { SecurityRequirementModel } from '../../services/models/SecurityRequirem
 import { linksCss } from '../Markdown/styled.elements';
 
 const ScopeName = styled.code`
-  font-size: ${props => props.theme.typography.code.fontSize};
-  font-family: ${props => props.theme.typography.code.fontFamily};
+  font-size: ${(props) => props.theme.typography.code.fontSize};
+  font-family: ${(props) => props.theme.typography.code.fontFamily};
   border: 1px solid ${({ theme }) => theme.colors.border.dark};
   margin: 0 3px;
   padding: 0.2em;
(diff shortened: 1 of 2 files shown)


RomanHotsiy:
This looks good! Thanks @valdirmendesdev 

I'm just not 100% sure on `Anonymous` term.

@cmeeren what do you think?

ian-zunderdorp:
If I can offer a suggestion: How about "Not applicable"?

cmeeren:
As demonstrated in the issue, it has to work together with authorization. So "Not applicable" won't work. I suggest "No authorization". Works in combination with another auth scheme, and describes exactly what it is.

cmeeren:
"None" could also work. That would avoid the awkward issue that is "authorization" vs. "authentication". The standard HTTP header is "Authorization", but it is a misnomer; the correct term is "Authentication". (Though ReDoc already displays this as "Authorization", so I'm not sure how much it matters.)

valdirmendesdev:
@cmeeren how about "not required"?!

RomanHotsiy:
All good suggestions here!

cc @adamaltman

cmeeren:
"Not required" could also work. Personally I would prefer "None", but not by much.

adamaltman (commented):
This is a good idea. 🎉 🥳 

In reference docs 2.0 (our premium product) we changed the name from "Authorization" to "Security". Why? Because security is in line with the name in the OpenAPI spec and can describe both authentication and authorization (what scopes are required for access). I think "None" still makes sense if that label changed to "Security" here as well.

One additional point, how about adding a test to prove it works and prevent regressions?

valdirmendesdev:
@adamaltman thanks for your suggestion. I will think how to test this behavior. Maybe rendering and testing the text of the result, what do you think?

RomanHotsiy:
@valdirmendesdev yes! You can use shallow render: https://github.com/Redocly/redoc/blob/master/src/components/__tests__/OneOfSchema.test.tsx#L34

valdirmendesdev:
@adamaltman I tried to write unit test to cover the expected behavior. Can you check if I am in the right way, please?

RomanHotsiy:
@valdirmendesdev that's great!

RomanHotsiy (approved):

RomanHotsiy:
Let me know if you're ready to merge or if you wanted to add anything else.

valdirmendesdev:
Thanks @RomanHotsiy, for now, I will not add anything else.

AlexVarchuk (approved):

PR #1699, opened 2021-07-20
feat: Add download file option

### LINKED ISSUE #672: `redoc-cli bundle` downloaded spec filename and format
When one bundles specs into a static html, the download button downloads a "swagger.json". It would be neat to be able to:

- Change the filename
- And make it possible to download the file as yml. The source being yml doesn't help in this case.

Filename changing is possible by adding the following to the template:
```
    <script>
        document.addEventListener("DOMContentLoaded", (event) => {
            document.getElementsByClassName('sc-gipzik')[0].download = 'renamed_spec.json';
        });
    </script>
```
but not as pretty and future-proof as a parameter.

zachwhaley:
## What/Why/How?

Allows the user to set a download file name and format for the API spec.

For example, `openapi.yaml`

## Reference

Fixes #672

## Testing

New tests written

## Check yourself

- [x] Code is linted
- [x] Tested

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines


RomanHotsiy:
How about just changing the default to openapi.yaml ?

zachwhaley:
@RomanHotsiy I believe there are still folks that would like the option to provide a JSON file as that is a valid format. 

I did not change the default just to keep backwards compatibility.

zcz3313:
@zachwhaley @RomanHotsiy can this PR be merged? our team needs this feature :)

src/services/RedocNormalizedOptions.ts
```diff
@@ -153,6 +155,20 @@ export class RedocNormalizedOptions {
...
+      const extname = path.extname(value);
+      if (extname === '.json' || extname === '.yaml') {
+        return value;
+      } else {
+        console.warn(`downloadFileName must be a JSON or YAML file.`);
+      }
+    }
+
+    // Default value
+    return 'swagger.json';
```
AlexVarchuk, line 169: I think you should check the version of the definition (to keep backward compatibility) or return the default name `openapi.yaml`.

AlexVarchuk (changes requested):

src/services/RedocNormalizedOptions.ts
RomanHotsiy:
```suggestion
    return 'openapi.json';
```

src/services/models/ApiInfo.ts
```diff
@@ -35,7 +41,13 @@ export class ApiInfoModel implements OpenAPIInfo {
     }
 
     if (IS_BROWSER && window.Blob && window.URL && window.URL.createObjectURL) {
-      const blob = new Blob([JSON.stringify(this.parser.spec, null, 2)], {
+      let specString: string;
+      if (path.extname(this.options.downloadFileName) === '.yaml') {
```
RomanHotsiy, line 45: we ignore the yaml dumper as it adds a lot of code to the bundle: https://github.com/Redocly/redoc/blob/master/webpack.config.ts#L116

Can we instead implement a downloadUrl option so you can provide any URL and don't have Redoc to serialise the definition.

RomanHotsiy (changes requested):
I would implement `downloadUrl` instead (see my comments). Let us plz know if you would be open to do it 🙌

src/services/models/ApiInfo.ts
zachwhaley: @RomanHotsiy I'm not sure I understand what a `downloadUrl` option would entail.  Are you suggesting users would provide their own URL for downloading a YAML API spec?  If so, what if the URL they provide does not contain the API spec or is inaccessible?

I had originally hoped that the original API spec file used to create the webpage would be available to use for download, but that did not seem to be the case or I did not know where to look for it.

RomanHotsiy: The original URL should work fine as long as you use URL and not use redoc-cli from a file on the filesystem (it should actually work like that but there may be a bug or something).

So I think it should work as below:

- if URL was used in Redoc settings - the download button should just point to this URL
- if a local file was used (e.g. redoc-cli) - it should use user-provided `downloadUrl`.
- if a local file was used and no `downloadUrl` was provided, then we just serialize it to json and use openapi.json as filename (JSON is free in terms of bundle size vs yaml).

What do you think?

>  If so, what if the URL they provide does not contain the API spec or is inaccessible?

The purpose of this option is to provide the URL for download button. Why should anybody provide a bad URL there?

zachwhaley: Sorry, I wasn't familiar with the multiple options for providing an API spec file; I believe our use case uses the redoc-cli.

So is there already an API spec URL that can be referenced in the code?  Or are you suggesting this be added; if so, would this URL be used to generate the HTML or would it just be for downloading an API spec file?

> Why should anybody provide a bad URL there?

I don't think it would be intentional, e.g. is it possible to use an internal or local company URL to generate the HTML pages?  If so, that URL might not be accessible to the wider internet.

adamdecaf:
Is there anything I can help with this to get it merged? Having a stable download URL would really help us auto-generate client code for multiple languages.

README.md
```diff
@@ -232,6 +232,7 @@ You can use all of the following options with standalone version on <redoc> tag
 * `expandResponses` - specify which responses to expand by default by response codes. Values should be passed as comma-separated list without spaces e.g. `expandResponses="200,201"`. Special value `"all"` expands all responses by default. Be careful: this option can slow-down documentation rendering time.
 * `maxDisplayedEnumValues` - display only specified number of enum values. hide rest values under spoiler.
 * `hideDownloadButton` - do not show "Download" spec button. **THIS DOESN'T MAKE YOUR SPEC PRIVATE**, it just hides the button.
+* `downloadFileName` - set file name and format of downloaded spec file, e.g. `openapi.yaml`, default `swagger.json`.
```
skadinna, line 235:
```suggestion
* `downloadFileName` - set a custom file name for the downloaded API definition file. The file is always in JSON format regardless of the file extension you set with this option. The default file name is `openapi.json`. **This option is supported only in redoc-cli**.
```

skadinna (commented):
@anastasiia-developer Hi Anastasiia, here's my suggestion for the description, please take a look :)

README.md
```diff
@@ -232,6 +232,7 @@ You can use all of the following options with standalone version on <redoc> tag
 * `expandResponses` - specify which responses to expand by default by response codes. Values should be passed as comma-separated list without spaces e.g. `expandResponses="200,201"`. Special value `"all"` expands all responses by default. Be careful: this option can slow-down documentation rendering time.
 * `maxDisplayedEnumValues` - display only specified number of enum values. hide rest values under spoiler.
 * `hideDownloadButton` - do not show "Download" spec button. **THIS DOESN'T MAKE YOUR SPEC PRIVATE**, it just hides the button.
+* `downloadFileName` - set file name of downloaded spec file, e.g. `swagger.json`, default `openapi.json`. Spec file in `JSON` format. (only `redoc-cli`)
```
skadinna, line 235:
```suggestion
* `downloadFileName` - set a custom file name for the downloaded API definition file. The file is always in JSON format regardless of the file extension you set with this option. The default file name is `openapi.json`. **This option is supported only in redoc-cli**.
```

README.md
```diff
@@ -232,6 +232,7 @@ You can use all of the following options with standalone version on <redoc> tag
 * `expandResponses` - specify which responses to expand by default by response codes. Values should be passed as comma-separated list without spaces e.g. `expandResponses="200,201"`. Special value `"all"` expands all responses by default. Be careful: this option can slow-down documentation rendering time.
 * `maxDisplayedEnumValues` - display only specified number of enum values. hide rest values under spoiler.
 * `hideDownloadButton` - do not show "Download" spec button. **THIS DOESN'T MAKE YOUR SPEC PRIVATE**, it just hides the button.
+* `downloadFileName` - set a custom file name for the downloaded API definition file. The file is always in JSON format regardless of the file extension you set with this option. The default file name is `openapi.json`. **This option is supported only in redoc-cli**.
```
anastasiia-developer, line 215:
```suggestion
* `downloadFileName` - set a custom file name for the downloaded API definition file. The file is always in JSON format regardless of the file extension you set with this option. The default file name is `openapi.json`. **This option is supported only in redoc-cli `build`**.
```

anastasiia-developer:
@zachwhaley  Hi! Sorry for the huge delay. Thank you for your contribution to redoc.

AlexVarchuk (approved):
Looks nice. 🎉

zachwhaley:
Thanks!  But removing the ability to download as yaml would mean https://github.com/Redocly/redoc/issues/672 can’t fully be closed.

RomanHotsiy:
> Thanks! But removing the ability to download as yaml would mean https://github.com/Redocly/redoc/issues/672 can’t fully be closed.

We will have to close it as WON'T FIX. We don't want to add a heavy yaml dump dependency which will increase a bundle size for this use-case.

Maybe, we will consider something comparable: if the source was yaml we can save yaml too.

RomanHotsiy (approved):

PR #1729, opened 2021-08-23
fix: improve action scripts

AlexVarchuk:
## Check yourself

- [ ] Code is linted
- [ ] Tested

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,12 +1,14 @@
-name: Unit Tests
+name: Tests
 
 on: [push]
 
 jobs:
-  build-and-unit:
+  build-and-test:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
     - run: npm ci
     - run: npm run bundle
     - run: npm test
+    - run: npm run e2e
+    - run: run run e2e-ci


.github/workflows/publish.yml
```diff
@@ -0,0 +1,136 @@
...
+          npm-${{ hashFiles('package-lock.json') }}
+          npm-
+    - run: npm ci
+    - run: npm run bundle
+  bundle-cli:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+      - uses: cd cli/
```
RomanHotsiy, line 29: no need to go inside cli folder

.github/workflows/publish.yml
```diff
@@ -0,0 +1,136 @@
...
+          key: npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            npm-${{ hashFiles('package-lock.json') }}
+            npm-
+      - name: Install dependencies
+        run: npm ci
+      - name: Bundle
+        run: npm run compile:cli
+  unit-tests:
+    needs: [bundle]
```
RomanHotsiy, line 43: unit tests can run without bundle

.github/workflows/publish.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Before deploy
+        run: npm run declarations
+      - name: Publish to NPM
+        run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: After acript
+        run: cat ./coverage/lcov.info | coveralls
+  publish-cli:
+    needs: [check-version-cli]
```
RomanHotsiy, line 118: I think it should have if statment

.github/workflows/tests.yml
```diff
@@ -0,0 +1,19 @@
...
+    - run: npm ci
+    - run: npm run bundle
+    - run: npm test
+  e2e-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: npm ci
+    - run: npm run bundle
+    - run: npm run e2e
```
RomanHotsiy, line 19: I would have e2e tests separately so we get faster feedback from unit tests

RomanHotsiy (changes requested):
Left a few changes requests.

The main issue here is I believe jobs do not share file systems so you have to use artefacts to share builds between jobs.

Also, I would split redoc-cli pipeline into separate action to simplify things.

Issue #1762, opened 2021-10-01
The number of items in the array in the array is incorrect


takumi-ricoh:
**Describe the bug**
If the maximum number of items in an array A1 is n1 and the maximum number of items in the array A2 in A1 is n2, the maximum number of items in A1 will be displayed as n2 after conversion.

**Expected behavior**
The maximum number of items in A1 is displayed as n1, and the maximum number of items in A2 is displayed as n2.

**Screenshots**
schema
![image](https://user-images.githubusercontent.com/32160810/135605456-48184fbe-fc1a-4902-8272-f3c9755ce566.png)

result
![image](https://user-images.githubusercontent.com/32160810/135605513-72297fd7-ac62-4ec8-911b-c817f663821f.png)

**Additional context**
OpenAPI v3.1
redoc-cli  0.12.3

PR #1308, opened 2020-06-25
add Array size info based on min max Items properties

mosheeshel:
Encountered [this issue](https://github.com/Redocly/redoc/issues/1131) in our spec, trying to fix according to suggestion in issue.

This is my first attempt to contribute to redoc (or any TypeScript/JavaScript project) - I hope I'm in the right direction, but I can't seem to build the project locally - tests are failing (on a clean master repo), so please be gentle.

If you can point me to how to add a test for this behavior, I'll gladly do that too.

Issue #1131, opened 2019-12-08
minItems and maxItems are not renders for array in request body


weismantelf:
Hello,

I am using the following definition to define an array of objects in the request body:

`               "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                  "type" : "array",
                                  "minItems" : 1,
                                  "maxItems" : 10,
                                  "items" : {
                                         "$ref": "#/components/schemas/gpsProfileUpdate"
                                   }
                            }
                        }
                    }
                }
 `

The minItems and maxItems were not rendered in the generated documentation.

![image](https://user-images.githubusercontent.com/58665051/70393838-ef8b0e00-19ee-11ea-9263-94fd62166a42.png)

RomanHotsiy:
Oh. I see.
Do you have any ideas on where to put this info?

m-mohr:
Either 

1. above "One Of" with respective labels 
    > Minimum number of items: 1
    > Maximum number of items: 10

  or

2. something inspired by programming languages where you specify the number of items in brackets (e.g. C: `array[5]` or Java: `new ArrayList(5)`):
    > Array (1 - 10) [
    > Array (no. of items: 1 - 10) [

Of course, if it's minItems = maxItems it would be better to just render 
> Array (5)

Just some ideas...

adamaltman:
I think option 1 would be the most clear to understand as a reader.

adamaltman:
Although, option 2 could be nice.  It may be more clear with the "items" word added to it:

range:
```
Array (1-5 items)
[
```

max == min (exact number)
```
Array (2 items)
```

max-only:
```
Array (5 or less items)
```

min-only:
```
Array (1 or more items)
```
It may be a pain though.  You have to worry about pluralizing "item" depending on the number of items.

AntonKozachuk:
Fixed by https://github.com/Redocly/redoc/pull/1308.

PR #1763, opened 2021-10-05
fix: The number of items in the array in the array is incorrect #1762

zalesky:
## What/Why/How?
closes https://github.com/Redocly/redoc/issues/1762
## Reference
the problem came from here https://github.com/Redocly/redoc/pull/1308 (description is here https://github.com/Redocly/redoc/issues/1131)

`minItems` and `maxItems` should be at the same level as `type: 'array'`, not at the `items` level

## Screenshots (optional)
Screenshot 2021-10-05 at 18 30 23

## Check yourself

- [ ] Code is linted
- [ ] Tested


AlexVarchuk:
Please check if it affects this [issue](https://github.com/Redocly/redoc/issues/1131).

Issue #1584, opened 2021-04-15
Webpack 5 support


villerdex:
Hi will there be any Webpack 5 support anytime soon?

rohit-gohri:
This is what you'll need to add to webpack config to fix build for webpack 5 or for use with latest docusaurus beta:

```js
import webpack from 'webpack';
import NodePolyfillPlugin from 'node-polyfill-webpack-plugin';

const webpackConfig = {
  plugins: [
    new webpack.DefinePlugin({
      'process.versions.node': JSON.stringify(
        process.versions.node || '0.0.0',
      ),
    }),
    new NodePolyfillPlugin({}),
  ],
};
```

If more errors pop up then this article was really helpful for debugging:
https://indepth.dev/posts/1166/this-will-make-you-more-efficient-at-debugging-webpack-unspecified-build-errors

pglezen:
I ran into this trouble when I attempted to upgrade my Docusaurus installation to the [beta version](https://docusaurus.io/blog/2021/05/12/announcing-docusaurus-two-beta), which internally updates its webpack to v5.

rohit-gohri:
> I ran into this trouble when I attempted to upgrade my Docusaurus installation to the [beta version](https://docusaurus.io/blog/2021/05/12/announcing-docusaurus-two-beta), which internally updates its webpack to v5.

@pglezen Off topic but I have made a redoc preset for docusaurus, might be useful to you: https://github.com/rohit-gohri/redocusaurus

theophile-scrive:
If I understand correctly, the beta version is unusable because it switches to webpack5 (which is not supported)?

Is there a way to constrain the webpack version as an end-user?

clarencedesmarais:
Tried the solution proposed by @rohit-gohri , unfortunately did not work for me with     "webpack": "5.62.1" , still getting : 
```
 Can't resolve 'fs' in ....
```

rohit-gohri:
> Tried the solution proposed by @rohit-gohri , unfortunately did not work for me with "webpack": "5.62.1" , still getting :
> 
> ```
>  Can't resolve 'fs' in path-to-app/node_modules/redoc/bundles
> ```

Updated my answer above to add this:

```
  resolve: {
    fallback: {
      fs: false,
    },
  },
```
 to fix this error.

pglezen:
> This is what you'll need to add to webpack config to fix build for webpack 5 or for use with latest docusaurus beta:
> 
> ...

[...]

PR #1796, opened 2021-11-19
fix: add browser build for webpack 5

AlexVarchuk:
## What/Why/How?
fixed https://github.com/Redocly/redoc/issues/1584
## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -32,10 +32,10 @@ const BANNER = `ReDoc - OpenAPI/Swagger-generated API Reference Documentation
   Version: ${VERSION}
   Repo: https://github.com/Redocly/redoc`;
 
-export default (env: { standalone?: boolean } = {}) => ({
+export default (env: { standalone?: boolean, browser?: boolean } = {}) => ({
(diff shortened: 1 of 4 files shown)


Oprysk (approved):

webpack.config.ts
```diff
@@ -47,13 +47,13 @@ export default (env: { standalone?: boolean } = {}) => ({
     fallback: {
       path: require.resolve('path-browserify'),
       http: false,
-      fs: false,
-      os: false,
+      fs: env.browser ? path.resolve(__dirname, 'src/empty.js') : false,
```
RomanHotsiy, line 50: I believe we don't need this conditional here, just keep it empty. It will be overwritten by `externalsPresets` anyways.

RomanHotsiy (approved):
LGTM 👍

CarltonHowell:
Looks like it's still busted in Webpack 5, getting the error in release candidate 59:
```
config.js:65 Uncaught (in promise) TypeError: path.join is not a function
```

RomanHotsiy:
Can you share a minimal reproducible sample?

Issue #1761, opened 2021-10-01
exclusiveMin shows [0...0]


takumi-ricoh:
**Describe the bug**
Display is strange when exclusive minimum is set

**Expected behavior**
shows >0

**Screenshots**
schema
![image](https://user-images.githubusercontent.com/32160810/135606553-828b9e94-fea3-46e0-9456-99ae7e5b3395.png)

result
![image](https://user-images.githubusercontent.com/32160810/135606583-2c34d227-fb74-468e-bf14-9edfde72e446.png)

**Additional context**
OpenAPI v3.1
redoc-cli 0.12.3

PR #1799, opened 2021-11-23
fix: exclusiveMin/Max shows incorect range

Oprysk:
## What/Why/How?
fix: https://github.com/Redocly/redoc/issues/1761
## Check yourself

- [x] Code is linted
- [x] Tested


src/utils/openapi.ts
```diff
@@ -448,16 +452,18 @@ export function humanizeConstraints(schema: OpenAPISchema): string[] {
...
-    let maximum = 0;
-    if (schema.minimum) minimum = schema.minimum;
-    if (typeof schema.exclusiveMinimum === 'number') minimum = minimum <= schema.exclusiveMinimum ? minimum : schema.exclusiveMinimum;
-
-    if (schema.maximum) maximum = schema.maximum;
-    if (typeof schema.exclusiveMaximum === 'number') maximum = maximum > schema.exclusiveMaximum ? maximum : schema.exclusiveMaximum;
-
-    numberRange = `[${minimum} .. ${maximum}]`
+  if (typeof schema.exclusiveMinimum === 'number' && typeof schema.exclusiveMaximum === 'number') {
+    numberRange = '[';
```
RomanHotsiy, line 456: Exclusive minimum means bigger, not bigger or equal. In math notation this corresponds to `(` vs `[`.
```suggestion
    numberRange = '(';
```

RomanHotsiy (changes requested):
This still has some issue:

```
exclusiveMinimum: 5
maximum: 10
```

Won't work correctly with this code.

Should show:

```
(5 .. 10]
```

src/utils/openapi.ts
Oprysk: thanks!
I've already cover this case

src/utils/openapi.ts
```diff
@@ -415,6 +419,29 @@ function humanizeRangeConstraint(
...
 }
 
+export function humanizeNumberRange(schema: OpenAPISchema): string | undefined {
+  const minimum =
+    typeof schema.exclusiveMinimum === 'number'
+      ? Math.min(schema.exclusiveMinimum, schema.minimum ?? Infinity)
+      : schema.minimum;
+  const maximum =
+    typeof schema.exclusiveMaximum === 'number'
+      ? Math.max(schema.exclusiveMaximum, schema.maximum ?? 0)
```
RomanHotsiy, line 429: It won't work for negative numbers. Would be good to add a test for this case maybe.
```suggestion
      ? Math.max(schema.exclusiveMaximum, schema.maximum ?? -Infinity)
```

RomanHotsiy (changes requested):
Left one small comment.

src/utils/openapi.ts
Oprysk: good point, thanks!

RomanHotsiy (approved):
🔥

AlexVarchuk (approved):

PR #1805, opened 2021-11-26
Use an option usePathInSidebar

kjhman21:
## What/Why/How?
We found an requirement that we want the path to be shown in the sidebar.
If this option is enabled, it displays a path in the sidebar instead of summary.

Usage:

```js
                <RedocStandalone
                spec={props.spec}
                options={{
                    scrollYOffset:'.navbar',
                    nativeScrollbars: true,
                    sideNavStyle: 'summary-only' or 'path-only'
                }}
```

## Screenshots (optional)
sideNavStyle == 'summary-only'
image

sideNavStyle == 'path-only'
image

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/services/MenuStore.ts
+++ b/src/services/MenuStore.ts
@@ -16,7 +16,7 @@ export interface IMenuItem {
   id: string;
   absoluteIdx?: number;
   name: string;
-  sidebarName: string;
+  sidebarLabel: string;
   description?: string;
   depth: number;
   active: boolean;
--- a/src/services/MenuStore.ts
+++ b/src/services/MenuStore.ts
@@ -16,6 +16,7 @@ export interface IMenuItem {
   id: string;
   absoluteIdx?: number;
   name: string;
+  sidebarName: string;
   description?: string;
   depth: number;
   active: boolean;
--- a/src/services/RedocNormalizedOptions.ts
+++ b/src/services/RedocNormalizedOptions.ts
@@ -22,7 +22,7 @@ export interface RedocRawOptions {
   disableSearch?: boolean | string;
   onlyRequiredInSamples?: boolean | string;
   showExtensions?: boolean | string | string[];
-  usePathInSidebar?: boolean;
+  sideNavStyle?: string | 'summary-only';
   hideSingleRequestSampleTab?: boolean | string;
   menuToggle?: boolean | string;
   jsonSampleExpandLevel?: number | string | 'all';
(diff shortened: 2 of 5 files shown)


AlexVarchuk (changes requested):
Hi @kjhman21 . Thanks for your contribution. Can you rename `usePathInSidebar` to `sideNavStyle`. It will be enum of `summary-only` and `path-only` with default value `summary-only`. We will add more options in the future.

kjhman21:
@AlexVarchuk Thanks for the quick review! Updated as you requested. Please take another look. :)

AlexVarchuk (approved):

src/services/RedocNormalizedOptions.ts
```diff
@@ -189,6 +206,7 @@ export class RedocNormalizedOptions {
   disableSearch: boolean;
   onlyRequiredInSamples: boolean;
   showExtensions: boolean | string[];
+  sideNavStyle: string;
```
RomanHotsiy, line 209: But would be awesome to use string enums here: https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

```suggestion
  sideNavStyle: 'summary-only' | 'path-only'
```

src/services/models/Operation.ts
```diff
@@ -104,6 +105,9 @@ export class OperationModel implements IMenuItem {
 
     this.name = getOperationSummary(operationSpec);
 
+    this.sidebarName = this.name;
+    if (options.sideNavStyle === 'path-only') this.sidebarName = this.path;
```
RomanHotsiy, line 109:
```suggestion
    this.sidebarName = options.sideNavStyle === 'path-only' ? this.path : this.name;
```

src/services/MenuStore.ts
```diff
@@ -16,6 +16,7 @@ export interface IMenuItem {
   id: string;
   absoluteIdx?: number;
   name: string;
+  sidebarName: string;
```
RomanHotsiy, line 19: I would rename it to `sidebarLabel`

RomanHotsiy (changes requested):
Thanks @kjhman21! It looks great! 💙

I left just a few tiny suggestions. Thanks in advance.

src/services/RedocNormalizedOptions.ts
kjhman21: @RomanHotsiy 
Actually, I didn't find string enums in the project. Could you suggest a good location to store enum definition?

kjhman21: Or, just your suggestion is enough?

RomanHotsiy: Hmm, keep it as my suggestion for now. We'll add enums later.

src/services/MenuStore.ts
kjhman21: applied as you suggested.

src/services/models/Operation.ts
kjhman21: applied as you suggested.

src/services/RedocNormalizedOptions.ts
kjhman21: @RomanHotsiy It looks ugly, so I applied string enums. Please take another look. Thanks!

RomanHotsiy (approved):
👏 Thanks @kjhman21. Great job!

@AlexVarchuk merge whenever you're ready.

Oprysk (approved):

kjhman21:
@RomanHotsiy @Oprysk @AlexVarchuk Thanks for approving this!

@AlexVarchuk When can this be merged? I am really excited to use this feature on my project!

Oprysk:
@kjhman21 Sorry for the delay, we are going to release it today!

Issue #1793, opened 2021-11-17
Does not serve local files


lukaslentner:
**Describe the bug**
My template is referencing files in the same folder (images, css, javascript) but the "serve"-webserver does not find them. I get back 404.

**Expected behavior**
The webserver should serve every file within the directory

AlexVarchuk:
Hi @lukaslentner,  please provide steps to reproduce and OpenAPI snippet.

lukaslentner:
[test.zip](https://github.com/Redocly/redoc/files/7555992/test.zip)

The index.html is created fine with the "bundle" command. But if you do "serve", then you do not see the image on the page and not as favicon...

PR #1810, opened 2021-11-29
fix: add handle for local files

AlexVarchuk:
## What/Why/How?
fixes https://github.com/Redocly/redoc/issues/1793
## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -6,7 +6,7 @@ import { ServerStyleSheet } from 'styled-components';
 
 import { compile } from 'handlebars';
 import { createServer, IncomingMessage, ServerResponse } from 'http';
-import { dirname, join, resolve } from 'path';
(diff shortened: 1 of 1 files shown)


RomanHotsiy (changes requested):
Hmmm, can we reuse openapi-cli preview-docs somehow here?

Maybe we should deprecate server command and ask people to use openapi-cli?

@AlexVarchuk what do you think?

AlexVarchuk:
> Hmmm, can we reuse openapi-cli preview-docs somehow here?
Maybe we should deprecate server command and ask people to use openapi-cli?

Good idea. I'll investigate it. Thanks

zalesky (approved):

RomanHotsiy (approved):

Issue #1794, opened 2021-11-18
Overrides user-provided discriminator values with object names


rmelasavvi:
**Describe the bug**

Happens if I define a "Vehicle", with a discriminator property of "vehicle_type", and then derive a "Car", with "vehicle_type: automobile".  The discriminator dropdowns and the sample requests ignore the values I've provided and use "Car" instead of the "automobile" value I've specified.

**Expected behavior**

If I define a Vehicle object with a discriminator propertyName of 'vehicle_type', and then in subclass Car I provide a value of 'automobile', then I expect the discriminator dropdowns, and the vehicle_type in the sample, to display "automobile", not "Car".

**Minimal reproducible OpenAPI snippet(if possible)**

```yaml
openapi: "3.0.1"
info:
  version: 1.0.0
  title: Vehicles
  description: Polymorphism example
servers:
  - url: https://api.example.com/vehicles/v1
paths:
  /vehicles:
    get:
      operationId: list_vehicles
      summary: List all vehicles
      responses:
        '200':
          description: An paged array of vehicles
          content:
            application/json:    
              schema:
                $ref: "#/components/schemas/Vehicle"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
  /vehicle:
    post:
      operationId: update_vehicle
      summary: create new vehicle
      tags:
        - vehicle 
      description: Create new vehicle
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Vehicle'
              
      responses:
        '200':
          description: Successfully created vehicle
          content:
            application/json:
              schema:
                type: object
                properties:
                  vehicle_id:
                    type: integer

              
  /vehicles/{id}:
    get:
      operationId: get_vehicle
      summary: Info for a specific vehicle
      tags:
        - vehicles
      parameters:
        - name: id
          in: path
          required: true
          description: The id of vehicle to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: "#/components/schemas/Car"
                  - $ref: "#/components/schemas/Plane"
                discriminator:
                  propertyName: type
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
    Vehicle:
      type: object
      required:
        - id
        - vehicle_type
      properties:
        id:
          type: integer
[...]

PR #1818, opened 2021-12-05
fix: Overrides user-provided discriminator values with object names (…

rmela:
…#1794)

## What/Why/How?

Fixes "Overrides user-provided discriminator values with object names"

## Reference

[issue 1794](https://github.com/Redocly/redoc/issues/1794)
[vehicles.yaml.txt](https://github.com/Redocly/redoc/files/7656573/vehicles.yaml.txt)

## Testing

Sorry, no new tests, though there should be.  But existing tests still pass.

## Check yourself

This pull request is more advisory/informational than a full pull request.  I did not write tests for the new functionality.

- [✔] Code is linted
- [✔] Tested
- [ x] All new/updated code is covered with tests

### DIFF
--- a/src/components/Schema/DiscriminatorDropdown.tsx
+++ b/src/components/Schema/DiscriminatorDropdown.tsx
@@ -33,7 +33,7 @@ export class DiscriminatorDropdown extends React.Component<{
 
     const options = parent.oneOf.map((subSchema, idx) => {
       return {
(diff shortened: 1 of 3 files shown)


RomanHotsiy:
I left details in the corresponding issue.

I have to close this PR as this adds extra complexity while solving the issue only for one specific case.

PR #1822, opened 2021-12-08
Replace bundlesize with size-limit

akx:
## What/Why/How?

[size-limit](https://www.npmjs.com/package/size-limit) is a more modern replacement for [bundlesize](https://www.npmjs.com/package/bundlesize).

In addition, bundlesize uses binary dependencies that are not required on modern Nodes.

## Check yourself

None of these really apply, so checking all of them. ✅ 

- [x] Code is linted
- [x] Tested


RomanHotsiy (approved):

RomanHotsiy:
Thanks @akx. It looks great!

Issue #1742, opened 2021-09-04
Unable to scroll on mobile


PythonCoderAS:
**Describe the bug**
Whener I try to scroll on mobile, it goes back to the top.

**Expected behavior**
I want to be able to scroll.

**Minimal reproducible OpenAPI snippet(if possible)**

Not my API, Link: https://api.mangadex.org

**Screenshots**
https://user-images.githubusercontent.com/13932583/132108543-95d408bc-2a32-408e-a3b9-4c8d40be56ec.MOV

skarzi:
It happens also on desktop (I am using Google Chrome 93.0.4577.63 on Linux)

ChristianHaase:
Can concur this happens on desktop as well (Chrome 92.0.4515.159 on Windows)

RomanHotsiy:
Can be related to https://github.com/Redocly/redoc/issues/1714?

skarzi:
> Can be related to #1714?

I can confirm that after disabling the flag mentioned in https://github.com/Redocly/redoc/issues/1714#issuecomment-897338497 scroll works like a charm in Chrome on Linux

PR #1823, opened 2021-12-09
Fix Chrome scrolling by using alternate scrollIntoViewIfNeeded

### LINKED ISSUE #1714: Experimental Web Platform features flag on chrome breaks scroll
This is a problem that I had first noticed on the Chrome Beta version, the sidebar menu is completely unusable, you can not click and it will randomly expand/collapse all by itself etc... 

At the time I had switched back to the normal stable chrome version and it worked as normal, however today I updated chrome to Version: `92.0.4515.131 (Official Build) (x86_64) on MAC` and it seems the Beta code has been pushed to stable.. which now breaks the redoc UI completely...

akx:
## What/Why/How?

Scrolling the documentation on Chrome seems to be broken in certain configurations (e.g. when Experimental Web Platform Features is enabled).

This seems to be caused by a faulty `scrollIntoViewIfNeeded` implementation (https://github.com/Redocly/redoc/issues/1714#issuecomment-897420205).

There already is a polyfill/implementation for the function in the codebase that works correctly in Chrome, so this changes things to always use that implementation instead of the possibly buggy native one.

## Reference

Fixes #1714

Refs #1742 (can't say it outright fixes it, didn't test on mobile)

## Testing

Seems to work on my Chrome!

## Check yourself

- [x] Code is linted
- [x] Tested
  - by hand 
  - no new code, no new tests

### DIFF
--- a/src/components/SideMenu/MenuItem.tsx
+++ b/src/components/SideMenu/MenuItem.tsx
@@ -7,6 +7,7 @@ import { shortenHTTPVerb } from '../../utils/openapi';
 import { MenuItems } from './MenuItems';
 import { MenuItemLabel, MenuItemLi, MenuItemTitle, OperationBadge } from './styled.elements';
 import { l } from '../../services/Labels';
+import { scrollIntoViewIfNeeded } from '../../utils';
 
 export interface MenuItemProps {
   item: IMenuItem;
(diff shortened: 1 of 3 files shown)


RomanHotsiy:
Good idea 👍 

We need to test it before merging though.

RomanHotsiy (approved):

Oprysk (approved):

PR #1838, opened 2021-12-16
feat: add try-live poc

cge:
This poc demonstrates integration of try-live functionality into the "Request samples" side panel area.

It does so via a new "Try" tab which exposes an embedded jsoneditor for payload editing. Currently only application/json mime type is supported. There is no schema validation for the poc, but it should not be hard to add.

Some small extra features include
 - server to use for try-live url selectable from info area
 - beginnings of request url param editing / styling
 - additional option params
   - tryLiveAccessToken
   - tryLiveSandboxServerIndex
   - tryLiveEditorTheme

Simply run with `npm start` and view at `http://localhost:9090/`.

This is just a humble poc in case it triggers ideas or interest. Thanks for a nice product


RomanHotsiy:
Hi @cge!

Thanks for your contribution. Unfortunately, this is not generic enough to merge it and also we have a try it in [our premium product](https://redoc.ly/reference/) so this is not on the priority roadmap for our OSS version at least for now.

cge:
Hi @RomanHotsiy!

Thanks for taking a look. Sure, I understand. It is certainly not generic enough for a merge, but shared to pique interest. Are there online demo's of the premium try it feature? It will be interesting to compare usability of a command-line approach vs inline. Whatever the case, I don't personally have time to follow up with this topic atm. Cheers!

RomanHotsiy:
@cge you can take a look at live example here: https://api-reference.rebilly.com/

PR #1841, opened 2021-12-24
fix: attachHeadingsDescriptions match headings incorrectly

Celend:
## What/Why/How?

if headings have same prefix, the description will mess up, see below:
```yaml
info:
  openapi: 3.1.0
  title: test
  version: v1
  description: |+
    # intro
    # modules
    ## user_module
    some text
   
    ## user
    some text2
```

then you can see:
image

how to fix?
just match heading to end of the line.

## Check yourself

- [x] Code is linted
- [x] Tested
- [ ] All new/updated code is covered with tests

PR #1845, opened 2021-12-30
fix: attachHeadingsDescriptions match headings incorrectly

Celend:
See https://github.com/Redocly/redoc/pull/1841

also fix an additional bug.


Oprysk:
@Celend Thank you for providing the details!
The additional context here is that this issue appears only with cli. I will take a look into this.

Oprysk (approved):

PR #1859, opened 2022-01-15
feat(#1251): Add file selector to demo application

DepickereSven:
## What/Why/How?
This PR should fix the request to add a file selector to the demo application see issue #1251 .

## Reference
See issue #1251 

## Testing
A GIF to demonstrate how it works
Upload local file

## Screenshots (optional)
afbeelding

The button to upload a file is hidden on smaller devices
afbeelding

## Check yourself

- [x] Code is linted
- [x] Tested


DepickereSven:
@Oprysk I noticed when opening a response you will get the following error 

> Uncaught TypeError: can't access property "expanded", this is undefined


This issue not only appears on my branch but also on master, are you aware of this issue?

demo/components/FileInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+    display: none;
+  }
+`;
+
+function GetUseRef(): RefObject<HTMLInputElement> {
+  return useRef<HTMLInputElement>(null);
+}
+
+const FileInput = props => {
+  const hiddenFileInput: RefObject<HTMLInputElement> = GetUseRef();
```
Oprysk, line 31: we can use it implicit `useRef<HTMLInputElement>(null)`, wrap func `GetUseRef` not needed

Oprysk:
@DepickereSven Thanks for your contribution!
All looks good, left a few comments.

demo/components/FileInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  function handleClick() {
+    if (hiddenFileInput && hiddenFileInput.current) {
+      hiddenFileInput.current.click();
+    }
+  }
+
+  function uploadFile(event) {
```
Oprysk, line 39: We would prefer to use regular func as a Component declaration and the arrow func as a handler inside the component.

Oprysk (approved):

PR #1867, opened 2022-01-22
fix: Response objects couldn't open

DepickereSven:
## What/Why/How?

When you clicked on a response object you would get the following error.

> Uncaught TypeError: can't access property "expanded", this is undefined

## Reference
No reference

## Testing
Added some cypress tests

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/e2e/integration/menu.e2e.ts
+++ b/e2e/integration/menu.e2e.ts
@@ -4,9 +4,7 @@ describe('Menu', () => {
   });
 
   it('should have valid items count', () => {
-    cy.get('.menu-content')
-      .find('li')
-      .should('have.length', 34);
+    cy.get('.menu-content').find('li').should('have.length', 34);
   });
 
   it('should sync active menu items while scroll', () => {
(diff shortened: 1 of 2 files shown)


e2e/integration/menu.e2e.ts
```diff
@@ -59,10 +55,25 @@ describe('Menu', () => {
...
+      .parent()
+      .find('div h3')
+      .should('have.text', 'Responses')
+      .parent()
+      .find('div:first button')
+      .click()
+      .should('have.attr', 'aria-expanded', 'true')
+      .parent()
+      .find('div h5')
+      .then($h5 => $h5[0].firstChild!.nodeValue!.trim())
```
DepickereSven, line 76: I wanted to get the value of the h5 element.
If I would just do `.should('have.text', 'Response Schema:')` on that h5 it would throw an error. Because the span values would also be in that comparison.

Oprysk:
@DepickereSven Thank you for the fix!

Oprysk (approved):

PR #1873, opened 2022-01-25
fix: expanded observable

AlexVarchuk:
## What/Why/How?
we have a problem with observable after `mobx` update https://github.com/Redocly/redoc/pull/1415. And collapse/collapse fields don't work.
This PR fixes it for the Field component. 

## Check yourself

- [ ] Code is linted
- [ ] Tested

### DIFF
--- a/src/components/Fields/Field.tsx
+++ b/src/components/Fields/Field.tsx
@@ -28,79 +28,79 @@ export interface FieldProps extends SchemaOptions {
   renderDiscriminatorSwitch?: (opts: FieldProps) => JSX.Element;
 }
 
-@observer
-export class Field extends React.Component<FieldProps> {
-  toggle = () => {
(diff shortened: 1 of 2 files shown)


Oprysk (approved):

src/components/Fields/Field.tsx
```diff
@@ -28,79 +28,79 @@ export interface FieldProps extends SchemaOptions {
...
+    field.expanded === undefined ? expandByDefault : field.expanded,
+  );
+  const toggle = () => {
+    if (field.expanded === undefined && expandByDefault) {
+      field.collapse();
     } else {
-      this.props.field.toggle();
+      field.toggle();
     }
+    setExpanded(field.expanded);
```
RomanHotsiy, line 51: Is this required? Mobx should work without extra state...

PR #1927, opened 2022-03-11
feat: Add support for displaying operationId in the sidebar

Rycochet:
## What/Why/How?

As with most companies that use OpenAPI we make use of the `operationId` to actually know what code to call through auto-generated code, which made this library unusable for generating documentation despite all the other benefits.

Simply added an option to the enum arguments to allow it to display the `operationId` - which is marked as optional in the spec, so provided a fallback to the `path` (the Summary is not useful in technical documentation), and added to the documentation.

No testing has been added for this - as I couldn't find any tests for the already existing functionality - the already existing tests have enough problems that my best comment is "this didn't break anything that wasn't already broken".

## Check yourself

- [x] Code is linted
- [ ] Tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ You can use all of the following options with the standalone version of the <red
 * `sideNavStyle` - can be specified in various ways:
   * **summary-only**: displays a summary in the sidebar navigation item. (**default**)
   * **path-only**: displays a path in the sidebar navigation item.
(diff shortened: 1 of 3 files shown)


Oprysk:
@Rycochet Thanks for your contribution!
This is a good idea but seems like it has not fully worked. So I've pushed one more commit to your forked repo to make it work properly 🙌 .

AlexVarchuk (approved):

PR #1934, opened 2022-03-17
chore: update openapi-core and yargs libs

AlexVarchuk:
## What/Why/How?
update `cli` libs for future `redoc-cli` fixes.

## Check yourself

- [x] Code is linted
- [ ] Tested


AlexVarchuk:
@RomanHotsiy What do you think, can we use lib with `Python-2.0` license?

Oprysk (approved):

PR #1974, opened 2022-04-28
fix: large text in examples value

### LINKED ISSUE #1958: Horizontal scrollbar for path and query parameters
**Describe the bug**
A horizontal scrollbar is missing for path and query parameters. So, large text is cut and not visible.

**Expected behavior**
A horizontal scrollbar should be shown for path and query parameters.

**Minimal reproducible OpenAPI snippet(if possible)**
Try with larger lines or words.

**Screenshots**
NA

**Additional context**
NA

anastasiia-developer:
## Reference
close #1958 
## Screenshots (optional)
Screenshot 2022-04-28 at 18 47 41
Screenshot 2022-04-28 at 18 50 36

## Check yourself

- [ ] Code is linted
- [ ] Tested

### DIFF
--- a/src/components/Fields/Examples.tsx
+++ b/src/components/Fields/Examples.tsx
@@ -31,6 +31,5 @@ export function Examples({ field }: { field: FieldModel }) {
 
 const ExamplesList = styled.ul`
   margin-top: 1em;
-  padding-left: 0;
-  list-style-position: inside;
+  padding-left: 17px;
 `;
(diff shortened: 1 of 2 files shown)


src/components/Fields/Examples.tsx
```diff
@@ -31,6 +31,5 @@ export function Examples({ field }: { field: FieldModel }) {
 
 const ExamplesList = styled.ul`
   margin-top: 1em;
-  padding-left: 0;
-  list-style-position: inside;
+  padding-left: 17px;
```
zalesky, line 34: Is it possible to use here something other than `17px;`? Like something from the theme or doesn't use at all

anastasiia-developer: I can use  `list-style-position: outside;` instead of `padding-left: 17px; `
and it will look like this
Screenshot 2022-04-29 at 11 16 12

AlexVarchuk (approved):

PR #2014, opened 2022-05-18
Link to property

### LINKED ISSUE #63: Linking to certain attributes
You did an cool job already via #56 which makes it already useful and nice for us and we will integrate it soon. I just had the idea that it would be useful to link directly to a property e.g. sometimes we get questions and we could then just say, please see the `allowed_vehicle` property and link it. Instead the customer now has to search manually through the docs to read everything and place it correctly.

fenyagg:
## What/Why/How?

Need the ability to make a reference to the property. 
I added property names at the end of the path with `|`. By default, all categories of this property are open.
Added background highlighting of this property.

## Reference
closes #63

## Screenshots (optional)
2dfb3c00a5

## Check yourself

- [ ] Code is linted
- [ ] Tested


fenyagg:
I will create a new mr after improvements and fixes

PR #2015, opened 2022-05-19
feat: show min properties max properties

### LINKED ISSUE #1343: Show minProperties/maxProperties constraints for additionalProperties
It seems like a bug or missing feature that Redoc-powered API docs don't show constraints for `additionalProperties`.

For example, when we indicate `minItems: 1` on an array, it shows the `non-empty` flag:

Screen Shot 2020-08-06 at 2 13 28 PM

But there's no such equivalent indicator on the nested `additionalProperties` field below, which uses `minProperties: 1`. Likewise, there's no indicator that it should be _exactly_ 1 in this case, because there's also `maxProperties: 1` as well.

Here's a snippet of the OpenAPI that produced the above screenshot

```
  SubscriptionRule:
    type: object
    required:
    - clauses
    properties:
      clauses:
        type: array
        minItems: 1
        description: |
          A rule can have multiple clauses, all of which must match for the rule to match.
          In other words, multiples clauses are `AND`'d together for evaluation.
        items:
          '$ref': '#/components/schemas/SubscriptionClause'

  SubscriptionClause:
    oneOf:
      ...
      - type: object
        description: Filter for alerts with properties that match the given name and value (exact string match)
        additionalProperties:
          type: string
          description: The name and value of the alert attribute to match
        minProperties: 1
        maxProperties: 1
        example:
          severity: info
```

Note: I'm sure there are complications, e.g., OpenAPI supports `minProperties`/`maxProperties` on objects with existing properties, not just the free-form `additionalProperties` field. So some design decisions will need to be made here. 😬

anastasiia-developer:
## What/Why/How?
close #1343 
## Screenshots (optional)
Screenshot 2022-05-18 at 20 38 13

Screenshot 2022-05-18 at 19 04 56
Screenshot 2022-05-18 at 19 05 23
Screenshot 2022-05-18 at 19 05 41
Screenshot 2022-05-18 at 19 07 43

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/demo/openapi.yaml
+++ b/demo/openapi.yaml
@@ -88,7 +88,7 @@ paths:
     parameters:
       - name: Accept-Language
         in: header
-        description: "The language you prefer for messages. Supported values are en-AU, en-CA, en-GB, en-US"
+        description: 'The language you prefer for messages. Supported values are en-AU, en-CA, en-GB, en-US'
         example: en-US
         required: false
         schema:
@@ -254,7 +254,7 @@ paths:
           required: false
           schema:
             type: string
-          example: "Bearer <TOKEN>"
+          example: 'Bearer <TOKEN>'
         - name: petId
           in: path
           description: Pet id to delete
@@ -401,6 +401,7 @@ paths:
             application/json:
               schema:
                 type: object
+                minProperties: 2
                 additionalProperties:
                   type: integer
                   format: int32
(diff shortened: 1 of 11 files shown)


zalesky (approved):

AlexVarchuk (approved):

clarencedesmarais:
I still see the `non-empty` field in `2.0.0-rc.76`, which is currently the latest version. 
image

Issue #2017, opened 2022-05-23
Lost functionality over weekend (worked on Friday) `oneOf` names not working and lost drop examples


cleverly87:
**Describe the bug**
oneOf functionality not displaying the tag names of the correct schema that it references to. 
Also the drop down JSON examples that had been hard coded are no long available in the third column.

**Expected behavior**

**Minimal reproducible OpenAPI snippet(if possible)**
I will copy a trail of the links as the API doc is huge.

PATH
"/localSolutions/GB/identitysearch": {
      "post": {
        "tags": [
          "GB Consumers and AML"
        ],
        "summary": "Submits a GB Consumer or AML Search",
        "description": "Submits a GB Consumer or AML depnding on the Product provided. Validates criteria for each individual search before submitting, and may return a list of error strings instead." ,
        "parameters": [
          {
            "$ref": "#/components/parameters/authToken"
          }
        ],
        "requestBody": {
            "$ref": "#/components/requestBodies/Identity" 
                              },
        

  "requestBodies": {
      "Identity": {
        "description": "Select the `products` tab for the request body for each of the products `AML` and `Consumer`",
          "content": {
              "application/json": {
                  "schema": {
                      "allOf": [
                        {
                         "description": "Identity Search",
                              "title": "Identity"
                        },
                        {
                          "$ref": "#/components/schemas/Connect.GBConsumer"
                        }
                      ]
      }
              
    }
  }               
 }
},

 "Connect.GBConsumer": {
        "type":	"object",       
          "description":	"Select product type for object parameters and example body.\n\n Options for an `AML Search` or `Consumer Search`",
               "oneOf": [
                {
                  "$ref" :	"#/components/schemas/AMLSearch"
                },
                {
                   "$ref": "#/components/schemas/consumerSearch"
                }
            ]        
          },      

      "AMLSearch": {
        "description":	"Search parameters for AML product",
        "properties":	{   
          "UniqueId":	{
            "type":	"string",
            "nullable":	true
          },
          "common": {
            "description": "Person search criteria",	
            "allOf":	[
            {
              "$ref":	"#/components/schemas/Connect.Identity.CommonSearchCriteria"
            }                      
          ]  
        }, 
          "consumer": {
            "description": "Second person search option",
            "$ref": "#/components/schemas/Connect.Identity.ConsumerSearchCriteria"
        },      
          "iDAml":	{
            "description":	"Reference to bank account detail search",
            "allOf":	[
              {
[...]

PR #2020, opened 2022-05-23
fix: merge allOf in correct order

AlexVarchuk:
## What/Why/How?
after releasing the conditional operators we have a problem with merging oneOf, anyOf, and other parts of the schema.

fixes https://github.com/Redocly/redoc/issues/2017

## Screenshots (optional)
Screenshot 2022-05-23 at 15 32 44

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/services/OpenAPIParser.ts
+++ b/src/services/OpenAPIParser.ts
@@ -274,6 +274,9 @@ export class OpenAPIParser {
         properties,
         items,
         required,
+        oneOf,
+        anyOf,
+        title,
         ...otherConstraints
       } = subSchema;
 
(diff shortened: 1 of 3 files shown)


Oprysk (approved):

src/services/OpenAPIParser.ts
```diff
@@ -324,9 +329,25 @@ export class OpenAPIParser {
         receiver.required = (receiver.required || []).concat(required);
       }
 
+      if (oneOf !== undefined) {
+        receiver.oneOf = [...(receiver?.oneOf || []), ...oneOf];
```
RomanHotsiy, line 333: This is not 100% valid. We can't just merge oneOf's. But this is not needed. We hoist oneOfs to the top-level before.

src/services/OpenAPIParser.ts
```diff
@@ -324,9 +329,25 @@ export class OpenAPIParser {
...
+      }
+
+      if (anyOf !== undefined) {
+        receiver.anyOf = [...(receiver?.oneOf || []), ...anyOf];
```
RomanHotsiy, line 337: Same here.

src/services/OpenAPIParser.ts
```diff
@@ -324,9 +329,25 @@ export class OpenAPIParser {
...
+      }
+
+      if (examples !== undefined) {
+        receiver.examples = [...(receiver.examples || []), ...examples];
```
RomanHotsiy, line 341: I think we shouldn't merge examples too. Let's just overwrite.

RomanHotsiy (changes requested):

anastasiia-developer (approved):

RomanHotsiy (approved):

PR #2028, opened 2022-05-31
Link to property

### LINKED ISSUE #63: Linking to certain attributes

fenyagg:
## What/Why/How?

Need the ability to make a reference to the property. 

Added links to properties. Added background highlighting of this property.
I added property names at the end of the path separeted with `|` (`%7C` after url encode). 
Discriminator and oneOf value separeted with `#`.
By default, all categories of this property are open.
If the reference to the property is in the response, the response will also be opened.

## Reference
closes #63

## Screenshots (optional)
add027d45a

## Check yourself

- [ ] Code is linted
- [ ] Tested


src/utils/helpers.ts
```diff
@@ -217,3 +217,6 @@ export function isArray(value: unknown): value is any[] {
 export function isBoolean(value: unknown): value is boolean {
   return typeof value === 'boolean';
 }
+
+// get location hash without #
+export const getLocationHash = () => decodeURI(location.hash).substr(1);
```
fenyagg, line 222: There may be problems with server side rendering here (`location`). I think it needs to be wrapped in a try catch.

fenyagg:
@AlexVarchuk @RomanHotsiy do you have any comments ?

adamaltman:
Hi @fenyagg thank you for your contribution. However, we don't review/accept contributions for the features we highlight in our premium offerings: https://github.com/Redocly/redoc#redoc-vs-reference-vs-portals

We will make this policy more clear. Sorry about that.

AlexVarchuk:
Hi @fenyagg. Thank you for your contribution.
Sorry again.

Issue #2048, opened 2022-06-10
Mis-nesting of aria roles on sidebar navigation


ryan-garrett:
The sidebar navigation uses `menuItem` but without the correct grouping parent role, instead using a `navigation` role.
The use of the `navigation` role on the `ul` also changes the role of that element from an implicit list to an explicit non-list, so rendering the implicit child list items without a correctly named parent.

<img width="705" alt="Screenshot 2022-06-10 at 14 49 35" src="https://user-images.githubusercontent.com/25553660/173080025-685a765d-3e7d-4f70-96f2-0c94618b36d7.png">

AlexVarchuk:
Hi @ryan-garrett, thank you for the issue.
Can you provide some OpenApi snippett for reproducing?

AlexVarchuk:
@ryan-garrett Please check our  [contribution guide](https://github.com/Redocly/redoc/blob/master/.github/CONTRIBUTING.md). And if you'll have some questions or improvements ideas about it feel free to write.

ryan-garrett:
@AlexVarchuk Sorry, forked and created a PR: https://github.com/Redocly/redoc/pull/2049

ryan-garrett:
Sorry @AlexVarchuk, my git messed up so I redone the pull request: https://github.com/Redocly/redoc/pull/2050

AlexVarchuk:
@ryan-garrett Thank you for your contribution. 🙌  Merged.

PR #2049, opened 2022-06-10
Fix: Mis-nesting of aria roles on side navigation

ryan-garrett:
The sidebar navigation uses menuItem but without the correct grouping parent role, instead using a navigation role.
The use of the navigation role on the ul also changes the role of that element from an implicit list to an explicit non-list, so rendering the implicit child list items without a correctly named parent.

Before:

Screenshot 2022-06-10 at 16 45 58

After:

Screenshot 2022-06-10 at 16 48 06

## Check yourself

- [X] Code is linted
- [X] Tested

### DIFF
--- a/cli/npm-shrinkwrap.json
+++ b/cli/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "redoc-cli",
-  "version": "0.13.15",
+  "version": "0.13.16",
   "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


AlexVarchuk:
@ryan-garrett please restore the `package-lock.json` file.

AlexVarchuk:
fixes https://github.com/Redocly/redoc/issues/2048

Issue #2042, opened 2022-06-03
Siblings of $ref do not override the referenced object


atruskot:
**Describe the bug**
Note: this is a duplicate of https://github.com/Redocly/redoc/issues/1727.  I'm not sure if this community prefers to reopen closed issues or create a new issue after closing the original. Feel free to close this issue if the preference is to reopen the original.

I'm still able to reproduce this issue in `redoc-cli@0.13.16`. Seeing the base description rather than the sibling/overridden one. Will see the sibling description if the base has no description. 

Worth noting, the issue isn't specific to description, any base property will render if it exists. So base examples can't be overridden either, for example. I'm not sure if the expectation is for the derived object to be arbitrarily override any base property or if Redoc is expected to only override specific properties (such as `description`). 

**Expected behavior**
When there are siblings of $ref, it is expected that the siblings override any properties that are also founded in the referenced object. 

**Minimal reproducible OpenAPI snippet(if possible)**
Described in https://github.com/Redocly/redoc/issues/1727.

AlexVarchuk:
The 3.1 include JSON Schema draft 2020-12 which allow all keywords siblings inside schema object. But the sibling keywords do not override the keywords in the referenced schema, they add to them. [Link to issue ](https://github.com/OAI/OpenAPI-Specification/issues/2744).

PR #2063, opened 2022-06-24
fix: merge reference for openapi 3.1

AlexVarchuk:
## What/Why/How?
we lost sibling functionality in version 2.0.0-rc.57-58. This PR fix it.
fix https://github.com/Redocly/redoc/issues/2042
## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/services/OpenAPIParser.ts
+++ b/src/services/OpenAPIParser.ts
@@ -1,6 +1,6 @@
 import { OpenAPIRef, OpenAPISchema, OpenAPISpec, Referenced } from '../types';
 
-import { isArray, isBoolean, IS_BROWSER } from '../utils/';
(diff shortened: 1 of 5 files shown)


src/services/OpenAPIParser.ts
```diff
@@ -173,7 +173,7 @@ export class OpenAPIParser {
       keys.some(k => k !== 'description' && k !== 'title' && k !== 'externalDocs')
     ) {
       return {
-        allOf: [rest, resolved],
+        allOf: [resolved, rest],
```
RomanHotsiy, line 176: Is it related to our recent changes in all of merging?

AlexVarchuk: Yes

Issue #213, opened 2017-02-27
Ellipsis class content is garbled


sarus:
I'm seeing garbled content for the `.ellipsis` class when the tree is collapsed.

![image](https://cloud.githubusercontent.com/assets/306319/23368214/198c3a3c-fcdb-11e6-9d59-325d16f8c1c9.png)

Inspector Content:

![image](https://cloud.githubusercontent.com/assets/306319/23368262/4070f386-fcdb-11e6-8a69-efbecb57d060.png)

Has anyone run into this issue?

Thanks!

sarus:
Just needed to add `UTF-8` charset as a `meta` tag in `head`

```
<meta charset="utf-8"/>
```

I would consider adding this to the sample `index.html` file in the README.

Thanks!

RomanGotsiy:
@sarus dup of https://github.com/Rebilly/ReDoc/issues/29

RomanGotsiy:
@sarus nice catch with `<meta charset="utf-8">`.
Adding to sample `index.html`
Thanks

raubel:
For information, it may not be enough to add `<meta charset="utf-8">` in the page. The response header can override this setting.
In my case, the `index.html` page is served by a Java servlet, and I need to set the response encoding explicitly (`response.setCharacterEncoding("UTF-8")`).

Issue #29, opened 2016-03-12
Strange character in contracted form


adamaltman:
Look at this before and after it's expanded.  Notice the strange character in the contracted version:

![image](https://cloud.githubusercontent.com/assets/1161871/13723210/c545db1c-e823-11e5-98da-445e7e313cd0.png)

![image](https://cloud.githubusercontent.com/assets/1161871/13723208/c10dd63a-e823-11e5-8a62-db7a758be885.png)

RomanGotsiy:
@adamaltman can you please provide more info: operating system, browser version, etc?

I can't reproduce this. Tried on IE11/Windows, Chrome/Mac and Safari/Mac.

adamaltman:
Linux (ubuntu) Chrome.  Oddly, I don't seem able to reproduce it now myself.  (At least I have the screenshots to prove I wasn't crazy.)

adamaltman:
I can reproduce this by running the deploy script on my local machine.

RomanGotsiy:
Are you able to reproduce it always? Or like time by time?
Once you reproduce it could you please copy content of redoc.js file and send it to me?

adamaltman:
The issue for me was the python-based simple http server which isn't utf-8 friendly.  This fixed the issue: http://stackoverflow.com/questions/15288891/how-can-i-serve-files-with-utf-8-encoding-using-python-simplehttpserver

PR #2130, opened 2022-08-11
fix: encoding issue in CDN responses

bparmentier:
## What/Why/How?

When downloading the redoc JS from redoc.ly CDN (e.g. https://cdn.redoc.ly/redoc/latest/bundles/redoc.standalone.js) the `Content-Type` header in the response is missing `'charset=utf-8'`, causing the file to be displayed with garbled characters.

I faced the issue in my project and found that it was indirectly reported multiple times over the years, as can be seen in https://github.com/Redocly/redoc/issues/29.

The solution proposed in https://github.com/Redocly/redoc/issues/213#issuecomment-282761418 is just a workaround and I'm actually not sure why it works for other people because according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-charset, "UTF-8 is the only valid encoding for HTML5 documents" anyway.

## Reference

- `aws-cli` doesn't automatically detect the charset when uploading files https://github.com/aws/aws-cli/issues/1346
- there was also a similar issue opened at https://github.com/jsdelivr/bootstrapcdn/issues/949

## Testing

Open the above URL in a browser and search for e.g. `.ellipsis:after`. The value of `content` should be displayed as 
image
instead of
image

### DIFF
--- a/scripts/publish-cdn.sh
+++ b/scripts/publish-cdn.sh
@@ -8,7 +8,9 @@ VERSION=$(node scripts/version.js)
 VERSION_TAG=v${VERSION:0:1}.x
 
 copy_to_s3 () {
-  aws s3 cp bundles "s3://redocly-cdn/redoc/$1/bundles" --recursive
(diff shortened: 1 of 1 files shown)


AlexVarchuk (approved):
Hi @bparmentier. Thank you a lot for your contribution 🙌

PR #2157, opened 2022-09-07
feat: add option to hide the example code when using the `SchemaDefinition` component

andre-stoesel:
## What/Why/How?

This change allows hiding the examples of a schema definition.

This is useful if the schema is not part of the API but an external resource like a configuration file. As these may have to be created in a different format, a `json` example would only confuse the user.

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/components/SchemaDefinition/SchemaDefinition.tsx
+++ b/src/components/SchemaDefinition/SchemaDefinition.tsx
@@ -14,6 +14,7 @@ export interface ObjectDescriptionProps {
   exampleRef?: string;
   showReadOnly?: boolean;
   showWriteOnly?: boolean;
+  showExample?: boolean;
   parser: OpenAPIParser;
   options: RedocNormalizedOptions;
 }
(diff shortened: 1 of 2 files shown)


AlexVarchuk:
Hi @andre-stoesel, thanks for your contribution. Can you add some examples of usage(definition) and screenshots for it?

andre-stoesel:
> Hi @andre-stoesel, thanks for your contribution. Can you add some examples of usage(definition) and screenshots for it?

An example is one of our web services. (Unfortunately the documentation is not finished and approved yet, so I can't link it here).
We document not only the API with redoc, but also the steps necessary for the installation and setup of the web service.
For this we simply create a tag with an appropriate description and use the `SchemaDefinition` component in markdown.
Similar as it is done in the `demo/openapi.yaml` file with the tag `pet_model`.

In our case the schema to be displayed corresponds to a configuration file in XML format. Since the `SchemaDefinition` component as far as I know only displays `json` as an example, we would rather hide this.

Changing the `pet_model` tag leads to the following result:

```yaml
  - name: pet_model
    x-displayName: The Pet Model
    description: |
      <SchemaDefinition schemaRef="#/components/schemas/Pet" showExample={false} />
```

image

As you can see, the example in the gray right side of the page is not displayed.

AlexVarchuk (approved):

PR #2159, opened 2022-09-09
feature: add unique className for components

ckoegel:
## What/Why/How?
Most components and other divs in the Redoc rendered API spec do not have unique `className`s. Users who want/need to override default Redoc CSS are required to use relative css because of this (e.g below). This leads to some object's CSS not being able to be changed without affecting other components on the page that also happen to be nested in the same way. This PR aims to give certain components unique class names to allow for more direct referencing, to solve the problem of clashing CSS, and improve code structure.

_Current CSS override in our code base for polymorphism buttons:_
```css
.RedocStandalone
  div[id^='operation']
  > div
  > div
  > div
  > div
  > div
  > div
  > button[class^='sc'] {
    background-color: var(--ifm-background-color);
}
```

## Testing
There are currently 3 tests in place that monitor component rendering, each of these tests has had their snapshots updated to include the new class names. All unit and e2e tests are passing.

There is also a new Markdown file created in the `docs` directory that lists the classNames that have been added.

## Check yourself

- [x] Code is linted
- [x] Tested


ckoegel:
@Oprysk Is there a time table for this to be looked at? It's currently blocking for us, and we would like to have it reviewed so we can move forward with either using an official release or a fork.

christopher-ayling:
@Oprysk This would be very helpful for us as well.

adamaltman:
@ckoegel I think this is incompatible with our plans. I'll confirm with @Oprysk soon.

ckoegel:
Hey @adamaltman @Oprysk , just wondering if you guys had taken a look at this. It's blocking for us and we'd really like to know if this is going to be merged soon or if we need to just fork the repo.

bpateldx:
@adamaltman @christopher-ayling Cameron has dedicated sometime to improve the support for styling by introducing `className` for key elements. The PR was opened more than three weeks back and it's still under review! This really hampers one's ability to improve respective docsite using **redoc**. I strongly suggest you both to review and approve the PR to help open source contributions to continue. With delays like this and no recommendations for alternative options there wouldn't be any other option but move away from this library.

RomanHotsiy:
Hi,

First of all, there is a reason why we don't have CSS classes. It will limit our ability to change/update redoc without introducing breaking changes.

With CSS classes almost every layout change becomes a breaking change and we still have a lot of different ideas/features on our roadmap which will require a lot of those changes.

We are considering multiple options to improve the styling of redoc, one of the options is to switch our JSON-based theme to an extensive set of CSS variables.
We are also considering adding classes with some opt-in flag that ensures people understand breaking changes risk.

> With delays like this and no recommendations for alternative options there wouldn't be any other option but move away from this library.

We need to have more internal discussions about this change. This is not something we are ready to merge in a swift. There are thousands of other people using Redoc and any change should be considered carefully.

We also have a bunch of other work to do so we can't allocate all of our time to this open-source and free library.
A simple alternative is to use a fork while we decide on this.

I would also recommend opening an issue before doing any work to better align with maintainers. Thanks!

AlexVarchuk:
Closed it. If merging happens, maintaining and changing code will become unbearable.

PR #2166, opened 2022-09-14
Update capitalization in README: ReDoc → Redoc

waldyrious:
Also add missing punctuation to warning, and change example CDN link to the now-released [2.0.0](https://github.com/Redocly/redoc/releases/tag/v2.0.0).

PR #1748, opened 2021-09-13
[GSoD]Updates readme with about section and organizes features list

HCloward:
## What/Why/How?
- Added back the TLDR section as requested in Slack
- Added an About Redoc section based on Ivana's original PR
- Added organization to features list
## Reference
[Project card](https://github.com/HCloward/redoc/projects/1#card-65203027)
## Testing
N/A
## Screenshots (optional)
N/A
## Check yourself
N/A
- [ ] Code is linted
- [ ] Tested
- [ ] All new/updated code is covered with tests

PR #2167, opened 2022-09-14
demo/index.tsx: Update to use new Redoc logo

waldyrious:
Currently the demo at https://redocly.github.io/redoc/ still uses the old logo:

image

This PR changes the demo code to use the new logo:

Redoc logo

Related changes:
- commit c4564de324d78c86e4436cf7ee0aa5b8c63f48ca from PR #1748 (/cc @HCloward)
- PR #2166
- Probably more, but I didn't dig further into the git history :smile:

### DIFF
--- a/demo/index.tsx
+++ b/demo/index.tsx
@@ -94,7 +94,7 @@ class DemoApp extends React.Component<
         <Heading>
           <a href=".">
             <Logo
-              src="https://github.com/Redocly/redoc/raw/main/docs/images/redoc-logo.png"
+              src="https://github.com/Redocly/redoc/raw/main/docs/images/redoc.png"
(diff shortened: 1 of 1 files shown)


waldyrious:
Note: this change should also be propagated to the demo bundle at https://github.com/Redocly/redoc/blob/gh-pages/redoc-demo.bundle.js.

I would also suggest that https://github.com/Redocly/redoc/blob/main/docs/images/redoc-logo.png be removed, or at least renamed to something like `redoc-logo-old.png`, to avoid confusion in the future.

AlexVarchuk:
Hi @waldyrious. Thank you a lot for the update 🔥 .

AlexVarchuk (approved):

PR #2207, opened 2022-11-16
Prevent loss of field descriptions when converting swagger specs by using refSiblings: 'preserve'

### LINKED ISSUE #2067: External dependency: swagger2openapi loses descriptions on complex objects
I've created an issue Mermade/oas-kit#597 over at **swagger2openapi** regarding an anomaly with converting field descriptions on swagger 2.0 documents.

I wanted to lodge it here as well in the advent of a fix.

The the descriptions on fields for complex objects are overwritten with the description of the object definition itself, and losing context for the user e.g. instead of displaying "The weight (in kg)..." and "The volume (in m3)..." of a reusable measurement object definition, it displays "A generic measurement." for both fields.

image

(I'll come back to update this as the issue progresses, if this is the right way to handle things?)

davetransom:
This change resolves #2067 for converting swagger 2.0 based specs to open-api 3.0.0.

## What/Why/How?
Previously, when using a swagger 2.0 file with ReDoc, the descriptions of any field using a complex type (referencing another schema) would show the description of the _type_, instead of the description of the _field_ itself.

## Reference
As per https://github.com/Mermade/oas-kit/issues/597#issuecomment-1228344557, `swagger2openapi` needs `refSiblings: 'preserve'` to bring additional fields into the converted open-api object. 

This change makes the description field available to the `fieldSchema` objects, but _it also needs_ to be merged into the field model to be shown in the UI.

1. Alters call to `swagger2openapi` to use refSiblings: 'preserve'
2. Alters `models\Field.ts` when _deciding_ on the description value to use, now also considers the `fieldSchema.description` in addition to the existing info or schema description.
3. Adds a new test to `FieldModel.test.ts` and a new swagger 2.0 fixture `complexFieldDescriptions.json` for testing.

Additionally, there are two options for consideration when choosing the description:
1. Use the `fieldSchema.description` verbatim if it's defined, or
2. Concatenate the field description with the schema description. However, this option might be considered unintended behaviour.

I'll update (clean) the code depending on your thoughts.

## Screenshots
If you would reuse a type, say "Measurement", for both weight and volume fields, the intended description is lost i.e. see the repetitive description in the first screenshot:
176603363-52afb20c-82dd-419f-a74f-30629e1e484c

where the _expected result_ is to show the explanation for the field:
image

## Checks

- [x] Code is linted
- [x] Tested

Final Notes:
- I did have some issues trying to get the vanilla _main_ branch to pass all tests before starting to make changes, so I've only focused on the tests applicable to this PR. I'm not sure if it's my local setup though...
- Some of the files in the base haven't had _prettier_ applied to them yet, so my changes to the `loadAndBundleSpec.ts` file would have had a lot of unrelated formatting changes to existing code. I opted out for that file only, so the diff is more representative of the change.
- I haven't created a lot of PRs before, so please let me know if I need to change/improve anything.

### DIFF
--- a/src/services/__tests__/fixtures/2.0/complexFieldDescriptions.json
+++ b/src/services/__tests__/fixtures/2.0/complexFieldDescriptions.json
@@ -1,0 +1,65 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "AA",
+    "version": "1.0"
+  },
+  "paths": {
+    "/test": {
+      "get": {
+        "operationId": "test",
+        "responses": {
+          "200": {
+            "description": "OK",
+            "schema": {
+              "$ref": "#/components/schemas/Box"
+            }
+          }
+        }
+      }
+    }
+  },
+  "definitions": {
+    "Measurement": {
+      "type": "object",
+      "description": "Reusable measurement type.",
+      "properties": {
+        "value": {
+          "format": "double",
+          "description": "The numeric value of the measurement.",
+          "type": "number"
+        },
+        "uom": {
+          "description": "The unit-of-measurement.  \r\ne.g. kg, m³",
+          "type": "string"
+        }
+      }
+    },
+    "Box": {
+      "type": "object",
+      "description": "Model which contains a reused type as a field e.g. a box with dimensions.",
+      "properties": {
+        "weight": {
+          "$ref": "#/definitions/Measurement",
+          "description": "The gross weight of the box and its contents (in kg)."
+        },
+        "volume": {
+          "$ref": "#/definitions/Measurement",
+          "description": "The volume of the box (in m3)."
+        },
+        "height": {
+          "$ref": "#/definitions/Measurement",
+          "description": "The height of the box (in mm)."
+        },
+        "width": {
+          "$ref": "#/definitions/Measurement",
+          "description": "The width of the box (in mm)."
+        },
+        "depth": {
+          "$ref": "#/definitions/Measurement",
+          "description": "The depth of the box (in mm)."
+        }
+      }
+    }
+  }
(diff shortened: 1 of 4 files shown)


AlexVarchuk:
Hi @davetransom. Thanks for your contribution. We appreciate that.
it looks good but relative to OpenApi 3.0 specification additional properties in the Reference Object shall be ignored. If we'll merge that it can affect the default behavior in the existing 3.0 definitions? I am not sure about that. 
@RomanHotsiy Do you have any thoughts about it?

davetransom:
I think that's a fair enough point @AlexVarchuk. 

This does only apply to the Swagger to Open API conversion (`convertSwagger2OpenAPI` method), so I don't think it will affect the default behaviour of the existing 3.0 definitions (non swagger related).

- Is the main concern that _other_ properties could creep in from the Swagger 2.0 format and bring the unintended change in Open API 3.0?
- Or do you mean any additional properties inserted into the Open API 3.0+ model would simply be ignored _after_ conversion from swagger 2.0?

While I'm testing in a fairly clean swagger 2.0 format (from in-production swagger) I guess it could get pretty wild. I could look at creating some more tests which try to break the conversion.

But perhaps this boils down to putting it behind an *opt-in* flag, perhaps in `RedocNormalizedOptions`, if we weren't entirely sure, and defer to users finding issues with it.


aside: In my original commit, I added a few comments to `models/Field.ts` where the description is merged which also might warrant some opt-in behaviour:
https://github.com/Redocly/redoc/pull/2207/files#diff-f89b60f0bf664d1d5e47242a68f88ee4fd5f81d4c847f89287341523ee5ad1cfR101-R126

Issue #447, opened 2018-03-30
Have an option to disable Redoc URL hash or to add a preffix to it


protetore:
We've been using ReDoc for some time, but now we have a requirement to embedded it inside one of our applications, but this application uses *HashRouter*, so when we access ReDoc, it's rewriting the URL removing the existing hash, which leads to an unknown route.

Example: `http://localhost/#/docs/` gets rewritten to `http://localhost/#/paths/~1something~1/get`.

It would be nice to have an option to *disable ReDoc URL hash* and, maybe, an option to add a URL *hash prefix*, to be added between the hash (#) and the ReDoc path ( `http://localhost/#/docs/paths/~1something~1/get`).

alairock:
Any updates or thoughts on this one? plz/thanks

brooksyd2:
Would also like to know if this is possible, particularly being able to prevent the hashing completely as this causes a lot of jumping around which is irritating.

luisfarzati:
My workaround was to do patch `MenuStore` like this:

```
import { RedocStandalone, MenuStore } from "redoc";

MenuStore.prototype.subscribe = function() {
  this._hashUnsubscribe = this.history.subscribe(this.updateOnHistory);
};
```

I'm using BrowserRouter and the hash change was triggering a re-render everytime, so scrolling down the page would take me all the way to the top again, rendering the page unusable.

ssharma-sumo:
any update on this ??

jrub:
Is there a way to remove the **tag** fragments like `#tag/` when using `x-traitTag`? 🙏

gish:
> My workaround was to do patch `MenuStore` like this:
> 
> ```js
> import { RedocStandalone, MenuStore } from "redoc";
> 
> MenuStore.prototype.subscribe = function() {
>   // this._unsubscribe = this.scroll.subscribe(this.updateOnScroll);
>   this._hashUnsubscribe = this.history.subscribe(this.updateOnHistory);
> };
> ```
> 
> I'm using BrowserRouter and the hash change was triggering a re-render everytime, so scrolling down the page would take me all the way to the top again, rendering the page unusable.

In TypeScript I had to do it this way, since the methods are private.
```js
MenuStore.prototype.subscribe = () => {
  // Do nothing
};
MenuStore.prototype.dispose = () => {
  // Do nothing
};
```

PR #2221, opened 2022-12-02
feat: add baseURL for redoc

Adol1111:
## What/Why/How?

We have a requirement to embedded it inside one of our applications, but this application uses HashRouter. ReDoc will rewrite the URL, which leads to an unknown route. So add a url prefix for document. 

fix: https://github.com/Redocly/redoc/issues/447

## Check yourself

- [ ] Code is linted
- [ ] Tested


lornajane:
Thank you for supporting Redoc and for your pull request. Adding the baseurl causes the tool to break in some settings, so we won't be accepting the pull request into the main project, in order to avoid causing problems for other users. Thanks again for being part of the project.

Issue #2141, opened 2022-08-23
Load Redocly icon without external CDN


seanking2919:
**Describe the problem to be solved**
On various sites using Redocly CLI > 0.11.0, the Redocly Attribution has a logo in it. The issue lies with said logo being pulled from a CDN, particularly cdn.redoc.ly. Many have different arguments as to why it should be loaded from an external CDN instead. However, loading it from an external CDN causes a page to try to connect to another domain. Hence, more loading time. In addition, it's known to cause numerous privacy concerns.

**Describe the solution you'd like**
I'd like for the Redocly icon to be loaded locally just like all the other images and resources.

adamaltman:
Thanks for the issue @seanking2919.

**Increase awareness**
We've heard from our customers that they wish it was easier to find us. Even those using our open source solution already weren't aware we offered more solutions such as a developer portal. We've worked to make the Redocly attribution remain out of the way for users but increase our brand awareness by being more visible than the prior attribution. 

**Difference from other images and resources**
The Redocly brand attribution has the Redocly brand in it. The other resources are specific to the version of Redoc installed (and may evolve over time). The Redocly brand may also evolve over time (let's take the case if we change our logo which is used as the icon), but this is the major difference: the attribution icon shouldn't be locked to a specific version of Redoc. In that case, we don't want to have the old brand icon floating around for years on end.

**Possible solution**
- Your proposed solution won't work for us, because we want to be able to rebrand quickly.
- We could potentially offer a configuration option, such as `hideRedoclyAttributionIcon`, which would not display the icon. 

However, before we add that option, let's get some more feedback. Adding yet another configuration option conflicts with another goal we have to reduce the number of configuration options. 🤣

seanking2919:
> Your proposed solution won't work for us, because we want to be able to rebrand quickly.

With all due respect, you could easily do a PR or a commit to the main branch and then release a new minor version (ie 0.11.1 or 0.11.2) if you really had to change branding quickly.

adamaltman:
> With all due respect, you could do a PR or a commit to the main branch and then release a new minor version (ie 0.11.1 or 0.11.2) if you really had to change branding quickly.

This is exactly what I mean. This doesn't work. It requires people to update. A quick PR on our end is not the challenge. It's updating 100,000 websites. How many of them run the latest version? How often do they update? How many haven't updated the version in over a month? over a year? over 5 years? (Yes, some big projects running Redoc haven't updated the version in over 5 years.)

[...]

PR #2222, opened 2022-12-05
Include Redoc logo in package

BenediktSeidl:
## What/Why/How?

Inline the SVG-Logo instead of loading it from a cdn.
We use redoc together with strict `Content-Security-Policy`-headers, and see an error `Content Security Policy: The page’s settings blocked the loading of a resource at https://cdn.redoc.ly/redoc/logo-mini.svg (“img-src”).` in the console.
This was resolved by inlining the SVG as already done with several other SVGs.

I first tried to load the svg image and inline it with the `asset/inline` feature of webpack.  Later on I saw that there are already svgs in the codebase. This is why there is the first commit removing the null-loader from `demo/webpack.config.ts`. I did not revert this change, as I think this is a good (but now unrelated) change. If you don't like it, I'm happy to remove it from this pull request.

## Reference

I implemented the svg similar to https://github.com/Redocly/redoc/blob/3d410b6002c656efa780254c9c45c6249f90bce1/src/components/StickySidebar/ChevronSvg.tsx#L29-L56

## Testing

* Seems to work in the playground, see screenshot `demo/webpack.config.ts` 

## Check yourself

- [x] Code is linted
- [x] Tested


lornajane:
Thanks @BenediktSeidl ! I'm very sorry for the slow response time on this pull request, we decided not to merge since the current setup works well for us and the majority of users. I'm closing it, but we appreciate having you as part of the project.

seanking2919:
> we decided not to merge since the current setup works well for us and the majority of users

On the contrary, look at the number of people who downvoted an official response from someone at Redocly to my previous call for loading the Redocly logo without the need for an external CDN: https://github.com/Redocly/redoc/issues/2141#issuecomment-1250393293

Issue #2225, opened 2022-12-06
WCAG 1.4.1 - Links are not clearly identifiable


dagfs:
**Describe the problem to be solved**

In markdown rendered in ReDoc, text-decoration is set to `auto`. This seems to be an allowed value, but I can not find it in the doc.

According to WCAG 1.4.1 this breaks the function can not be communicated only trough color rule.

<img width="397" alt="image" src="https://user-images.githubusercontent.com/26817802/205935524-278b5ae6-664c-46f7-99cc-46ac2fc41c3a.png">

**Describe the solution you'd like**
I suggest setting the default for the default template to `text-decoration: underline;` as to promote accessability.

**Describe alternatives you've considered**
- I considered just changing this in our theme. But I think it is better if the default theme has this change.

**Additional context**

This issue is related to my previous issue #2223 as part of the same improvments

This issue relates to the tracking WCAG issue https://github.com/Redocly/redoc/issues/1279

RomanHotsiy:
I don't think that using underline as default is a good solution.

I'm closing it.

PR #2226, opened 2022-12-06
Change theme default value as to meet WCAG

dagfs:
## What/Why/How?

This solves: https://github.com/Redocly/redoc/issues/2225

In summary: The current default value for text decoration for links breaks WCAG 1.4.1, a level A fault. WCAG 1.4.1 says function should not be communicated by color alone. 

## Testing

As per the problems mentioned in #2225 I have not been able to run the test properly. The changes are minimal and mainly visual.

## Screenshots (optional)
Before
image

After
image

## Check yourself

- [x] Code is linted
- [ ] Tested


RomanHotsiy:
Sorry, I don't think that changing the default to underline is a good option here. Closing it.

Thanks for contribution!

Issue #2198, opened 2022-11-04
Missing variables for some elements


ops1ops:
Im customizing redoc via theme and found that some elements dont use variables, but they need to

https://github.com/Redocly/redoc/blob/main/src/common-elements/fields-layout.ts#L208

https://github.com/Redocly/redoc/blob/main/src/common-elements/headers.ts#L47

Issue #560, opened 2018-07-11
Theme doesn't cover all colors


anton-fedorov:
Hello,

I played with Theme settings in order to set custom color scheme. While it worked perfectly for objects, that you could configure, some of them seem to be not configurable.
I took the default theme object from https://github.com/Rebilly/ReDoc/blob/master/src/theme.ts, and changed all colors to blue and yellow (so that it is contrast) like below:

```js
{
    colors: {
        main: '#DFC350',
        success: '#DFC350',
        redirect: '#DFC350',
        error: '#DFC350',
        info: '#DFC350',
        text: '#DFC350',
        code: '#DFC350',
        codeBg: 'rgba(223, 195, 80, 1)',
        warning: '#DFC350',
        http: {
            get: '#DFC350',
            post: '#DFC350',
            put: '#DFC350',
            options: '#DFC350',
            patch: '#DFC350',
            delete: '#DFC350',
            basic: '#DFC350',
            link: '#DFC350',
            head: '#DFC350'
        }
    },
    schemaView: {
        linesColor: '#DFC350',
        typeNameColor: '#DFC350',
        typeTitleColor: '#DFC350',
        requireLabelColor: '#DFC350',
        nestingSpacing: '#DFC350'
    },
    links: {
        color: '#DFC350',
        visited: '#DFC350',
        hover: '#DFC350'
    },
    menu: {
        backgroundColor: '#445AFF'
    },
    rightPanel: {
        backgroundColor: '#DFC350',
        textColor: '#445AFF'
    }
}
```

On the screenshots below you can see a lot of changed colors, but things that were not configurable remained in default colors and are marked with arrows.

Could you please add options to set colors for those objects?

Thanks!

<img width="1400" alt="redoc1" src="https://user-images.githubusercontent.com/1927974/42560072-851bda30-84fe-11e8-84d4-52a03ce3df13.png">
<img width="1054" alt="redoc2" src="https://user-images.githubusercontent.com/1927974/42560078-87d40036-84fe-11e8-9714-50d7213efb0a.png">

Api Spec in case you need to see what are the defined elements:
[api_spec.txt](https://github.com/Rebilly/ReDoc/files/2183666/api_spec.txt)

RomanGotsiy:
Thanks for detailed issue report.

I am aware of this and working to cover all the colors.
PRs are welcome!

lezhnev74:
[...]

PR #2259, opened 2023-01-31
feat: Adds theme property to change background color of nested properties

david-mk-lawrence:
## What/Why/How?

This adds a property to the theme object to allow changing the background color of nested properties.

Partially addresses #560 and #2198 

For example

```javascript
<RedocStandalone
    options={{
        schema: {
            nestedPropertiesBackground: "#333333"
        }
    }}
/>
```

## Testing

Can test with a schema that contains an object that is at least 3 levels deep. For example,

```yml
responses:
  "200":
    content:
      application/json:
        schema:
          type: object
          properties:
            data:
              type: object
              properties:
                attributes:
                  type: object
                  properties: # <-- background color of this element should match the new property
                    field:
                      type: string
```

## Check yourself

- [x] Code is linted
- [x] Tested


AlexVarchuk:
Hi @david-mk-lawrence. Thank you for the contribution. 🙌 
I am not sure to add more options to Redoc. Let's wait, if people really need it, we'll think how to implement it better way.

RomanHotsiy:
How is it different from `nestedBackground`? I think it should use global background color if we have it or let's add it if we don't.

kipe:
@AlexVarchuk @RomanHotsiy 

I would assume that this change would allow changing the color in this situation:
image

Currently I can't at least find a way to alter that, and as you can see, it both breaks the theme and makes description text unreadable.

PR #2320, opened 2023-05-17
feat: remove redoc-cli

AlexVarchuk:
## What/Why/How?
As redoc-cli is no-support we should remove it from main branch. For now we use [@redocly/cli](https://github.com/Redocly/redocly-cli).
We saved it in `redoc-cli` branch for future. 
## Check yourself

- [x] Code is linted
- [x] Tested


lornajane:
Thanks! I think this makes things clearer for users

PR #2326, opened 2023-05-25
docs: updates the redocly cli command prefix from openapi to redocly

HCloward:
## What/Why/How?

The redocly CLI command prefix has been updated from openapi to redocly so this documentation needs to be updated to reflect that update.

## Check yourself

- [x] Code is linted


RomanHotsiy (approved):

lornajane (approved):
Thanks for picking up and fixing this!

PR #2397, opened 2023-08-30
Style RefreshToken URL as <code> in the OAuthFlow

ckszabi:
Represent refreshToken URL in a similar fashion like Token URL is styled, for consistency in rendering.

## What/Why/How?
When an OAuth security scheme definition is rendered the refreshToken URL is presented as plain text, but it should be a (non-clickable) URL, rendered via `<code>` tag, as the tokenURL is rendered.
One line of template is changed

## Screenshots (optional)
the wrong rendering:
image

better rendering:
image

## Check yourself

- [x] Code is linted
- [x] Tested


AlexVarchuk (approved):

AlexVarchuk:
Hi @ckszabi, thank you for the contribution 🙌

PR #2401, opened 2023-09-06
chore: update react-tabs version 3 -> 4 to be compatible with react@18

tatomyr:
## What/Why/How?

Support for React 18 has been recently added in Redoc. However, the `react-tabs@3` only supports React 16 or 17. This causes issues when installing React 18 together with Redoc. 

## Reference

See the issue: https://github.com/Redocly/redocly-cli/issues/1240

## Check yourself

- [ ] Code is linted
- [x] Tested


tatomyr:
Closed in favor of this: https://github.com/Redocly/redoc/pull/2404

PR #2442, opened 2023-10-26
added support for extensionsHook for ShelfIcon

pranitkhadilkar7:
## What/Why/How?
 - This PR adds support to change style in ShelfIcon using extensionsHook

## Check yourself

- [ ] Code is linted
- [ ] Tested


pranitkhadilkar7:
@RomanHotsiy , can you please review this PR? We need this change to do some styling on the arrow icon

lornajane:
Can you say a bit more about your use case? It's not clear to me why we'd want to support this, or how our users would know what it is and how to use it.

lornajane:
Closing, it's not clear how this would benefit our users.

Issue #2408, opened 2023-09-07
Accessibility: skipped heading levels (h2 -> h5)


kburk1997:
**Describe the bug**
1. Go to [Redocly Demo](https://redocly.github.io/redoc/#tag/pet/operation/newPet)
2. Run WAVE Chrome Extension
3. Observe heading levels are skipped, h2 -> h5 (see Screenshots)

**Expected behavior**
Heading levels should not be skipped.
**Minimal reproducible OpenAPI snippet(if possible)**

**Screenshots**
![Screenshot of WAVE run on Redocly demo with heading levels highlighted. Heading order is h2 -> h5](https://github.com/Redocly/redoc/assets/11298269/b262b0ed-9c42-4f79-8dd9-f8d55c6f73e8)
**Additional context**
[Deque University - Heading levels should only increase by one](https://dequeuniversity.com/rules/axe/4.4/heading-order)

PR #2443, opened 2023-10-27
Headings Hiearchy Fixed

shubhamt619:
## What/Why/How?
Headings Hierarchy was missing tags.

## Reference
Refer to #2408 

## Tests
All test cases passed, because I updated the test cases where needed.

## Check yourself

- [X] Code is linted
- [X] Tested
- [X] All new/updated code is covered with tests


src/common-elements/headers.ts
```diff
@@ -41,7 +47,7 @@ export const RightPanelHeader = styled.h3`
   ${extensionsHook('RightPanelHeader')};
 `;
 
-export const UnderlinedHeader = styled.h5`
+export const UnderlinedHeader = styled.h4`
```
AlexVarchuk, line 50: why do we need to change h5->h4?

src/components/Operation/Operation.tsx
```diff
@@ -36,7 +36,7 @@ export const Operation = observer(({ operation }: OperationProps): JSX.Element =
       {options => (
         <Row {...{ [SECTION_ATTR]: operation.operationHash }} id={operation.operationHash}>
           <MiddlePanel>
-            <H2>
+            <H3>
```
AlexVarchuk, line 39: Is it necessary to change level of heading in this case?

AlexVarchuk (changes requested):
Hi @shubhamt619, thank you for your contribution.
I have a couple of questions/clarifications

PR #2466, opened 2023-12-15
Create docker-image.yml

lukins-cz:
## Check yourself

- [ ] Code is linted
- [ ] Tested


lornajane:
Hi @lukins-cz could you add a description to the pull request to explain what the change is and why? You can use Redoc with docker already, the instructions are here: https://redocly.com/docs/cli/installation/

PR #2490, opened 2024-02-12
docs: Fix a link that couldn't be checked as it reaches outside the Redoc project

lornajane:
## What/Why/How?

Tiny link fix! In the context of the Redoc docs, the link checker can't check links that point to the locations in the wider marketing site. The link checker in that project caught the error, but it must be fixed here.

## Reference

broken build (Redocly users only) https://github.com/Redocly/marketing-site-portal/actions/runs/7712790410/job/21021123579?pr=1190

## Check yourself

- [x] Code is linted
- [x] Tested


HCloward (approved):
Checked link in preview build - works great!

PR #2506, opened 2024-03-28
feat: Update API examples list to include Museum API

### LINKED ISSUE #2473: Update sample to use Museum API

Our [Redoc demo](https://redocly.github.io/redoc/) defaults to PetStore. Update it to use our [Museum API](https://github.com/redocly/museum-openapi-example) instead.

Demo should load Museum API as the default demo. Let's add Petstore to the list though.

**Describe alternatives you've considered**

Keep a competitor's resource as the default.

lornajane:
## What/Why/How?

We have more modern examples that Redoc works pretty well with - so let's use them.

Fixes #2473 

## Check yourself

- [x] Code is linted
- [x] Tested


AlexVarchuk (changes requested):
Left few comments.
In general, I really like it. Finally, we will have a new example. 🙌 

What do you think about adding more complicated examples (discriminator, oneOf, markdown with SchemaDefinition, if/then/else, additional properties, etc)? Not all of that but a few useful cases, I believe, it will make the example more fancy and useful. 🚀

lornajane:
This is useful feedback, could you please add it over on the museum project https://github.com/Redocly/museum-openapi-example which is the source of truth for this file? We can bring a new version in to this project when we have one.

The purpose of this pull request is to use our existing Museum API as the main demo on https://github.com/Redocly/museum-openapi-example. I hope that you can help me to package it and push the correct content to the `gh-pages` branch. We need this for an upcoming event, and the issue had been open quite a while.

We can publish new Museum API versions as that project improves and evolves - but we need the demo updated with whatever we have, as soon as we can.

AlexVarchuk (approved):
Very weird but in Museum repo it works well. Fixed. 
Today I am going to deploy gh-pages

PR #2533, opened 2024-04-28
Allow showing SchemaDefinition as child

DuckyC:
## What/Why/How?

When using SchemaDefinition to show a ref that is a child of an object with discriminator, sometimes you want to show that exact child, and not the parent. 

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/components/SchemaDefinition/SchemaDefinition.tsx
+++ b/src/components/SchemaDefinition/SchemaDefinition.tsx
@@ -15,6 +15,7 @@ export interface ObjectDescriptionProps {
   showReadOnly?: boolean;
   showWriteOnly?: boolean;
   showExample?: boolean;
+  showAsChild?: boolean;
   parser: OpenAPIParser;
   options: RedocNormalizedOptions;
 }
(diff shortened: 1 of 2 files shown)


src/components/SchemaDefinition/SchemaDefinition.tsx
```diff
@@ -39,15 +40,10 @@ export class SchemaDefinition extends React.PureComponent<ObjectDescriptionProps
...
     if (!this._mediaModel) {
-      this._mediaModel = new MediaTypeModel(
-        parser,
-        'json',
-        false,
-        SchemaDefinition.getMediaType(schemaRef, exampleRef),
-        options,
-      );
+      const info = SchemaDefinition.getMediaType(schemaRef, exampleRef);
+      this._mediaModel = new MediaTypeModel(parser, 'json', false, info, options, showAsChild);
```
DuckyC, line 46: Added `showAsChild` but removing newlines was prettier

DuckyC:
Would love some guidance on how a test could be added to this 🙏

PR #2597, opened 2024-09-23
nnDelete config directoryspan

zori26:
## Check yourself

- [ ] Code is linted
- [ ] Tested


AlexVarchuk:
Hi @zori26, I don't understand what is going on. Seems, PR doesn't make sense.
Please add details on what it should fix.

PR #2600, opened 2024-10-13
feat: update pattern styling (#2196)

### LINKED ISSUE #2196: Render patterns (regexes) in a monospaced font
When using `pattern`, it currently renders like this (the rightmost part in green):

image

Since `pattern` is a regex, i.e. essentially code, it should be rendered in a monospaced font. In the above example, this would also make it much clearer that the pattern allows spaces. (The space before the final `-` is very thin and easy to miss if just scanning quickly over the pattern to get a grip on the allowed characters. As it is currently rendered, I would feel the urge to copy-paste it to a monospaced editor to make sure.)

miqh:
Hello,

This is a small change that adjusts the styling for patterns being displayed inline for a field.

I've included before and after screenshots using a part of the pet store example.

image

image

---

Closes #2196

### DIFF
--- a/src/common-elements/fields.ts
+++ b/src/common-elements/fields.ts
@@ -97,9 +97,11 @@ export const RecursiveLabel = styled(FieldLabel)`
 
 export const PatternLabel = styled(FieldLabel)`
   color: #0e7c86;
+  font-family: ${props => props.theme.typography.code.fontFamily};
+  font-size: 12px;
   &::before,
   &::after {
-    font-weight: bold;
+    content: ' ';
   }
 `;
 
--- a/src/components/__tests__/__snapshots__/FieldDetails.test.tsx.snap
+++ b/src/components/__tests__/__snapshots__/FieldDetails.test.tsx.snap
@@ -163,7 +163,7 @@ exports[`FieldDetailsComponent renders correctly when field items have string ty
     >
       [
       <span
-        class="sc-kpDqfm sc-eDPEul cGRfjn erJHow"
+        class="sc-kpDqfm sc-eDPEul cGRfjn cCKYVD"
       >
         ^see regex[0-9]$
       </span>


src/common-elements/fields.ts
```diff
@@ -97,9 +97,11 @@ export const RecursiveLabel = styled(FieldLabel)`
 
 export const PatternLabel = styled(FieldLabel)`
   color: #0e7c86;
+  font-family: ${props => props.theme.typography.code.fontFamily};
+  font-size: 12px;
   &::before,
   &::after {
-    font-weight: bold;
```
miqh, line 102: I doubt this style alone was doing anything resourceful, so I've removed it. I could be wrong though.

src/common-elements/fields.ts
```diff
@@ -97,9 +97,11 @@ export const RecursiveLabel = styled(FieldLabel)`
...
+    content: ' ';
```
miqh, line 104: The intention here is to ensure the pattern string is adequately spaced from adjacent inline elements.

Without it, there can be cases (see the before screenshot) in which there's no spacing at all.

src/components/__tests__/__snapshots__/FieldDetails.test.tsx.snap
```diff
@@ -163,7 +163,7 @@ exports[`FieldDetailsComponent renders correctly when field items have string ty
     >
       [
       <span
-        class="sc-kpDqfm sc-eDPEul cGRfjn erJHow"
+        class="sc-kpDqfm sc-eDPEul cGRfjn cCKYVD"
```
miqh, line 166: Had to update this snapshot since the style adjustment changes the output here.

AlexVarchuk:
Hi @miqh, thank you for the contribution. Please merge it when you are ready 🙌

PR #2623, opened 2024-11-26
feat: add virtualization (configurable)

### LINKED ISSUE #2615: Feature Request: Virtualization
When the OpenAPI gets very large (190K+ lines of code in the OpenAPI), it causes Redoc performance to drop significantly. In average consumes 1.2GB of the browser's memory and it took 1.2 minute for the First Contentful Paint (FCP).

One of the potential solutions I analyzed is Virtualization. Currently it would render the whole HTML page, which is quite good for small OpenAPIs, but could hurt huge API specifications.

By having Virtualization, I believe it will improve Redoc significantly on larger API Documentations as the browser will only render what is currently user is seeing.

audipasuatmadi:
## What/Why/How?
Projects with very large APIs often experience significant lag and frequent crashes when using Redoc. To address this issue, virtualization has been introduced. Virtualization allows elements to be rendered only when they are near the user's viewport, significantly reducing memory consumption for very large API documentation.

Virtualization is made optional because not every project would benefit from enabling it by default. It is advantageous only when the data is large, as it introduces an additional overhead of calculating the virtualization. When the data is not substantial, the drawbacks of this overhead outweigh the benefits.

**Benchmark result**
Using edge profiler, each recorded in 20 seconds in an OpenAPI docs with 192,076 (190K+) lines
image

## Reference
Closes #2615 

## Tests
Introduced extra unit tests. Results of running all the tests:
image

## Screenshots (optional)

https://github.com/user-attachments/assets/ecda9bec-7abc-469e-aaac-b86eea30cfe4

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/components/Virtualization/VirtualizedContent.tsx
+++ b/src/components/Virtualization/VirtualizedContent.tsx
@@ -9,6 +9,8 @@ type VirtualizedContentProps = {
   menu: MenuStore;
 };
 
+const ESTIMATED_EACH_API_HEIGHT_PX = 1000;
+
 /**
  * VirtualizedContent optimizes the rendering of API documentation in Redoc by virtualizing the content.
  *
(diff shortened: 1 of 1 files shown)


NGanip (approved):

alfinm01 (approved):

AlexVarchuk (changes requested):
Hi @audipasuatmadi thank you for the contribution to Redoc. 
I investigated your solutions, but it totally broke navigation and scrolls in the documentation, we cannot merge it as it is. Please improve it. Thanks.

PR #2633, opened 2024-12-19
feat: add option `sideNavLayout`

dmosberger:
I'm wondering if you'd consider adding a feature that allows making the sideNav menu more compact when there are end points sharing the same path prefix.  The below is a proof  of concept.  It seems to work quite well with `redocly preview-docs`, but I'm not very familiar with `Redoc` so may be missing things.

--------------------

Add support for theme option `sideNavLayout`.  If set to `factored`, the side-nav menu will group items that share a common path prefix.

For example, if the default layout for side-nav looked like this:

```
	/config        >
        /config/alert  >
        /config/bacnet >
        /config/modbus >
        /log           >
```

Then the factored side-nav would look like this:

```
	/config        >
        /log           >
```

and after expanding /config, you'd see:

```
	/config        v
          GET /config
          PUT /config
          …/alert      >
          …/bacnet     >
          …/modbus     >
        /log           >
```

Note that this commit also updates the styling to show a vertical line to the left of all side-bar `<ul>` lists that are not at the root level. This probably should be configurable.

## Check yourself

- [ ✓ ] Code is linted
- [ x ] Tested
- [ x ] All new/updated code is covered with tests

### DIFF
--- a/src/services/MenuBuilder.ts
+++ b/src/services/MenuBuilder.ts
@@ -8,6 +8,15 @@ import type { ContentItemModel, TagGroup, TagInfo, TagsInfoMap } from './types';
 
 export const GROUP_DEPTH = 0;
 
+function appendChildren(parent: ContentItemModel, children: ContentItemModel[], prefix: string) {
+  for (const child of MenuBuilder.factorByPrefix(children)) {
+    if (child.sidebarLabel.startsWith(prefix)) {
+      child.sidebarLabel = '…' + child.sidebarLabel.slice(prefix.length - 1);
+    }
+    parent.items.push(child);
+  }
+}
+
 export class MenuBuilder {
   /**
    * Builds page content structure based on tags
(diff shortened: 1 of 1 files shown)


AlexVarchuk (changes requested):
Hi @dmosberger, Thank you for your contribution.
Our navbar is built from `tags` and `x-tagGroups`. You can put the exact tag to each operation, and it works as you expected. If you want to do that programmatically you can use `@redocly/cli` and create a custom plugin to make changes. [Here](https://redocly.com/docs/cli/custom-plugins/custom-decorators) is an example of how to create a custom plugin.
I believe this way is more clear instead of adding a new maintainable option.

dmosberger:
Alex, thanks for your response.

It's entirely possible I'm missing something, but when I read the description [here](https://redocly.com/docs-legacy/api-reference-docs/specification-extensions/x-tag-groups), it left me with the understanding that `x-tagGroups` are rendered flat.  In my example, the menu would be rendered like this:

```
       CONFIG TAG GROUP
        /config        >
        /config/alert  >
        /config/bacnet >
        /config/modbus >
       LOG TAG GROUP
        /log           >
```

Also, it appears that tag groups are not collapsible and cannot be nested.

You can see a live example of the nested menu we need [here](https://webapi.egauge.net/).  Note how when you expand the `/config` node, other grouped nodes appear, e.g, `/config/alert`, which in turn contains group `/config/alert/custom`.

Scutua (approved):

PR #2646, opened 2025-01-14
docs: update options for future major release

AlexVarchuk:
## What/Why/How?
update documentation options for future major release

## Reference
https://github.com/Redocly/redoc/pull/2647
## Check yourself

- [ ] Code is linted
- [ ] Tested

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -218,7 +218,7 @@ _Default: false_
 If set to `true`, the API definition is considered untrusted and all HTML/Markdown is sanitized to prevent XSS.
 
 ## Theme settings
-Change styles for the API documentation page **Supported in Redoc CE 2.x **.
+Change styles for the API documentation page **Supported in Redoc CE 2.x**.
 * `spacing`
   * `unit`: 5 # main spacing unit used in autocomputed theme values later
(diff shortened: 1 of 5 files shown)


adamaltman (commented):
I would if we should move them to a "Deprecated options" section of that page for the interim.

docs/config.md
```diff
@@ -26,111 +26,158 @@ Sets the minimum amount of characters that need to be typed into the search dial
...
-Hides the Security panel section.
+### showExtensions
 
-### hideSingleRequestSampleTab
+Shows specification extensions ('x-' fields). Extensions used by Redoc are ignored. The value can be boolean or an array of strings with names of extensions to display. When used as boolean and set to `true`, all specification extensions are shown.
 
-Hides the request sample tab for requests with only one sample.
+### sanitize
 
-### htmlTemplate
```
adamaltman, line 77: Don't we still need this option? This is commonly used.

AlexVarchuk: Yes, you are right, it is very common. But `htmlTemplate` is not a Redoc CE option, it relates to `redocly-cli`. We moved it when was deprecated `redoc-cli` and moved code of build command to `redocly-cli`

adamaltman (approved):

PR #2654, opened 2025-02-07
feat: add keyboard navigation support to JsonViewer component

### LINKED ISSUE #2619: Accessibility: Keyboard focus is not coming on the horizontal scrollbar of Request and Response Samples
**Describe the bug**
When the request or response sample text overflows the width of the container, a horizontal scrollbar appears. 
If the request/response does not contains a "focusable" element, then the user can't use the left and right arrow keys to view the rest of the code in that section.

**Expected behavior**
If the request/response sections present a scrollbar, they should be keyboard accessible.

**How to reproduce**
1. On Chrome (not sure about other browsers).
2. Go to example page, `POST Buy museum tickets` method: https://redocly.github.io/redoc/#tag/Tickets/operation/buyMuseumTickets
3. Resize browser window to a small width, so that the "Request samples" of the method shows a horizontal scrollbar.
4. Try tabbing to get to the "Request samples" and use right/left arrow keys to focus. It is not possible.

**Screenshots**
Image

**Additional context**
For this bug, the request/response text does not need to have any focusable elements (like JSON expand/collapse buttons). The method referred in "how to reproduce" is an example that should display the bug.

**Solution** 
Add `tabindex="0"` to the element that has the scrollbar.

ImBIOS:
## What/Why/How?
Fix #2619

## Reference
_null_

## Tests

Tested manually and unit test

## Screenshots (optional)
_loom might make more sense if necessary_

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/components/JsonViewer/JsonViewer.tsx
+++ b/src/components/JsonViewer/JsonViewer.tsx
@@ -41,6 +41,7 @@ const Json = (props: JsonProps) => {
         <OptionsContext.Consumer>
           {options => (
             <PrismDiv
+              tabIndex={0}
               className={props.className}
               // tslint:disable-next-line
               ref={node => setNode(node!)}
--- a/src/components/__tests__/JsonViewer.tsx
+++ b/src/components/__tests__/JsonViewer.tsx
@@ -2,6 +2,7 @@
 
 import { mount, ReactWrapper } from 'enzyme';
 import * as React from 'react';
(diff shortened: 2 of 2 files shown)


RomanHotsiy:
Thanks for the contribution!

We'll review it soon.

AlexVarchuk (approved):

PR #2655, opened 2025-02-07
fix: enhance accessibility for menu items with keyboard support

### LINKED ISSUE #2621: TOC sections cannot be toggled with keyboard
**Describe the bug**
The sections on the left table of contents cannot be toggled (open/close) via keyboard, only with a mouse.

Image

**Expected behavior**
The sections should be expanded/collapsed when they have keyboard focus and expand/collapse with the space bar.

**How to reproduce**
1. Go to example page on Chrome: https://redocly.github.io/redoc/
2. Tab to get to a section like "Upcoming Events".
3. Press space to try to open/close - nothing happens.

**Recommendations**
- Ensure there is a `button` role either by making the control a `<button>` element or adding `role="button"` to the element
- Make the control focusable either by making it a `<button>` element or adding `tabindex="0"`
- Ensure the button can be activated with the ENTER and SPACE keys
- Add an accessible name to the button by adding an `aria-label` to the `button` or `<div role="button">`. Additionally, ensure the value updates when the double arrow icon updates to an "X" icon that closes the disclosure.
- Add the `aria-expanded` attribute to the button. When content is expanded, `aria-expanded="true"`, when it is collapsed, `aria-expanded="false"`

ImBIOS:
## What/Why/How?
Fix #2621

## Reference
_null_

## Tests
- tested manually and locally.
- unit test not included due to the component need a lot of code to write to jsut write test for this one function, can be added as additional task.

## Screenshots (optional)
_loom with keyboard stroke visualizer might makes more sense_

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/src/components/SideMenu/MenuItem.tsx
+++ b/src/components/SideMenu/MenuItem.tsx
@@ -2,14 +2,14 @@ import { observer } from 'mobx-react';
 import * as React from 'react';
 
 import { ShelfIcon } from '../../common-elements/shelfs';
+import type { IMenuItem } from '../../services';
 import { OperationModel } from '../../services';
-import { shortenHTTPVerb } from '../../utils/openapi';
-import { MenuItems } from './MenuItems';
-import { MenuItemLabel, MenuItemLi, MenuItemTitle, OperationBadge } from './styled.elements';
 import { l } from '../../services/Labels';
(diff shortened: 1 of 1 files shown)


AlexVarchuk (approved):
@ImBIOS  Thank you for the contribution.
Looks good

PR #2697, opened 2025-06-05
feat(i18n): use translate=no attribute in code sensitive renderings

### LINKED ISSUE #2696: Add translate=no attribute or notranslate classes in non-descriptive rendered fields

When using an external translation service, such as localizejs.com or Google Translate, the services can not distinguish which elements of a Redoc generated page should be translated and which should be left as-is.

A common convention is to use a `<code>` element, use the W3C defined `translate="no"` attribute, or apply a specific class to any element, such as `notranslate` (honored by Google Translate and LocalizeJS.com).

Redoc should add this common attribute (`translate="no"`) or class (`notranslate` to content derived from API spec fields other than `summary` and `description` (others may apply).

**Describe alternatives you've considered**

I do not know if these approaches are universal among client-side translation tools. This setting may need to be configurable by users, including the special class name.  For a minimal change, I would error on the side of honoring the W3C defined attribute (by default) and giving users an opt-out flag.

**Additional context**

* https://cloud.google.com/translate/troubleshooting
* https://www.w3.org/International/questions/qa-translate-flag
* https://developers.localizejs.com/docs/using-html-syntax-to-prepare-phrases#attribute-notranslate-or-translateno

Relates to #367

displague:
## What/Why/How?

Adds `translate="no"`, a W3C standard HTML attribute described at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/translate, and described in the issue linked below. This hints localization tools (Google Translate and others) to avoid translation of data/type sensitive sections of Redoc rendered OpenAPI specs. Many sites may exclude their Redoc pages from translation today because the outcome would be unpredictable. By adding `translate="no"` the translators and translation tools will know to leave these sections untranslated.

## Reference

Fixes #2696 

## Tests

- `npm test` (unit + license check): 290 tests, 29 suites, 52 snapshots — all passing.
- `npm run lint`: clean (one pre-existing warning in `src/components/RedocStandalone.tsx`, untouched by this PR).
- `npm run prettier`: no changes.
- `npm run ts-check`: clean.
- Added `src/components/__tests__/translateNo.test.tsx`, which asserts the attribute in both directions: code-sensitive surfaces are marked, and Redoc-authored prose is not. Each assertion was verified to be load-bearing by stripping the attribute from the sources under test and confirming the tests fail.
- Added `src/services/__tests__/mdtranslate.test.ts` for the Markdown renderer, including a check that surrounding prose stays translatable.
- Manually reviewed the rendered spec (`museum.yaml`) via `npm start`.

`npm run e2e` was not run: `@redocly/workerize-loader` is listed in `package.json` but missing from the local install, so the standalone bundle does not boot and all 26 specs fail. I confirmed this is pre-existing and unrelated by building and running e2e on an unmodified `upstream/main`, which fails identically. The Markdown change wraps `<pre>`/`<code>`, so I checked `Markdown/styled.elements.tsx` for `>`/sibling selectors — all rules are descendant-based (`code {}`, `pre {}`, `pre code {}`), so styling is unaffected.

## Check yourself

- [x] Code is linted
- [x] Tested


src/components/ApiInfo/ApiInfo.tsx
```diff
@@ -82,11 +82,12 @@ export class ApiInfo extends React.Component<ApiInfoProps> {
...
                     <DownloadButton
-                      download={downloadFileName || true}
-                      target="_blank"
-                      href={url}
-                      rel="noreferrer"
                       key={url}
+                      href={url}
+                      download={downloadFileName}
+                      target="_blank"
+                      rel="noopener noreferrer"
```
displague: I didn't ask for `noopener`.  (nor did I ask for fields to be reordered)

displague: The Download button is very likely safe to be translated and this hunk should be removed.

src/utils/jsonToHtml.ts
```diff
@@ -27,11 +27,11 @@ function stringifyStringLiteral(str: string) {
 }
 
 function decorateWithSpan(value, className) {
-  return '<span class="' + className + '">' + htmlEncode(value) + '</span>';
+  return '<span class="' + className + '" translate="no">' + htmlEncode(value) + '</span>';
```
displague, line 30: I haven't looked at everywhere this is called, I imagine decorateWithSpan will need an argument to determine if translation can be permitted.

src/utils/jsonToHtml.ts
```diff
@@ -27,11 +27,11 @@ function stringifyStringLiteral(str: string) {
...
 }
 
 function punctuation(val) {
-  return '<span class="token punctuation">' + val + '</span>';
+  return '<span class="token punctuation" translate="no">' + val + '</span>';
```
displague, line 34: Punctuation will likely always be open to translation (since I doubt it is used adjacent to code-sensitive fields derived from the spec).

src/utils/jsonToHtml.ts
```diff
@@ -106,7 +106,7 @@ function objectToHTML(json, maxExpandLevel: number) {
     const key = keys[i];
     hasContents = true;
     output += '<li><div class="hoverable ' + collapsed + '">';
-    output += '<span class="property token string">"' + htmlEncode(key) + '"</span>: ';
+    output += '<span class="property token string" translate="no">"' + htmlEncode(key) + '"</span>: ';
```
displague, line 109: I'm not sure where/how this is used. If these are OAS3 schema properties or values, then they should not be translated.

displague (commented):
This PR is in draft and not considered to be complete. I believe I've seen `<div>` used for some text elements in Redoc renderings. In the prompt for this change I hinted that likely candidates for change would be `<span>` elements.

I don't believe `<code>` needs to be decorated the same way, but we could easily add that if we can not assume translation services block those elements.

src/utils/jsonToHtml.ts
displague: Seems `decorateWithSpan` is only called by `valueToHTML`, which is only called with schema fields that should be translate=no decorated.

src/utils/jsonToHtml.ts
displague: Seems objectToHTML is from schema objects, code.

src/utils/jsonToHtml.ts
displague: This is used for object punctuation (`[{,}]`) and should not be translated.

displague:
I see the `translate=no` appearing in the right-side panel, the JSON example blocks, but I'm not seeing it in the left side where the descriptions are. I would appreciate any pointers.

src/components/ApiInfo/ApiInfo.tsx
displague: Resolved — this hunk is gone. `ApiInfo.tsx` now only adds `translate="no"` to the version `<span>`; the `DownloadButton` props are back to their original values and order (`download={downloadFileName || true}`, `rel="noreferrer"`).

src/utils/jsonToHtml.ts
displague: `decorateWithSpan` is only reached via `valueToHTML` on schema values, so no extra argument is needed. Left as `translate="no"`.

src/utils/jsonToHtml.ts
displague: kept as `translate="no"` — this only emits structural JSON punctuation (`[`, `{`, `,`, `}`, `]`).

src/utils/jsonToHtml.ts
displague: Kept as `translate="no"` — these are JSON/schema property keys.

displague:
> I'm not seeing it in the left side where the descriptions are

Found it — rebased onto latest `main` and pushed.

The left/description column renders through the shared styled elements in `src/common-elements/fields.ts`, which had no coverage at all, so nothing there was getting the attribute. Rather than annotating each call site, I applied it centrally with `.attrs({ translate: 'no' })`:

- `TypeName` — and `TypeFormat`, which aliases it, so `<format>` / `<contentEncoding>` / `<contentMediaType>` are covered too
- `PatternLabel` — regex patterns
- `ExampleValue` — `default:`, `example:`, `const:`, and array-form enum values
- `ExtensionValue` — inherits from `ExampleValue`
- `ConstraintItem` — the inline attribute from the first pass is now redundant and was dropped

Plus `<td translate="no">` for the value column of the `x-enumDescriptions` table.

Left translatable on purpose: `TypePrefix` ("Array of"), `TypeTitle` (author-written schema title), `RecursiveLabel`, required/deprecated labels, and descriptions.

Two other fixes in this push:

- `Field.tsx` had an unrequested restructure that moved the property name and parent-name prefix *out* of the clickable `<button>` — same class of edit as the `ApiInfo.tsx` one flagged above. Reverted to the original structure; it now only adds the attribute.
- `.redoc-json` now carries `translate="no"` on the container, rather than relying solely on per-token spans.

276 tests pass; 5 snapshots updated to record the new attribute.

displague:
Did a full pass over every rendering surface rather than just the `<span` sites the original prompt targeted. Two corrections and a set of gaps.

### One false positive, removed

`ConstraintItem` should **not** be `translate="no"`. Its content comes from `humanizeConstraints()` in `src/utils/openapi.ts`, which emits Redoc-authored English prose — `non-empty`, `unique`, `>= 5 characters`, `[ 1 .. 10 ] items`, `multiple of 3`, `decimal places <= 2`. That is UI text that *should* be translated. It was marked in the first pass; reverted.

### The largest gap: `StyledPre`

`src/common-elements/samples.tsx`. Every Prism-highlighted sample renders through it — all **request samples** (curl / JS / Python), non-JSON payload and response examples, and external-example error output. Nothing in the first pass touched it; only JSON examples were covered, because those route through `jsonToHTML` instead.

One implementation note: `StyledPre` and `StyledMarkdownBlock` both derive from `PrismDiv`, so the attribute goes on `StyledPre`. Putting it on the shared base would have suppressed translation of every description in the document.

### Other gaps closed

- **Security identifiers** — `SecurityDetails.tsx`, `OAuthFlow.tsx` and `RequiredScopesRow.tsx` used bare `<code>` for the API key parameter name, HTTP auth scheme, bearer format, OAuth flow type, authorization / token / refresh / connect URLs, and scope names. Plus the `ScopeName` styled element.
- **HTTP verbs in the sidebar and callback titles** — `Endpoint`'s `HttpVerb` was covered but `OperationBadge` was not. Only the verb usages are marked: the `type="hook"` badge renders `l('webhook')` when `showWebhookVerb` is off, so it is marked conditionally.
- **Callback names** — OpenAPI runtime expressions such as `{$request.body#/callbackUrl}`.
- **`fieldParentsName` prefix in `Field.tsx`** — parent property names were rendered as bare text *outside* the `.property-name` span, so for `parent.child` the prefix was translated while the leaf was not.
- Response status codes (`Code`), media type labels (`MimeLabel`, both definitions), and `x-` extension names.
- **Code inside descriptions** — overrode `renderer.code` / `renderer.codespan` in `MarkdownRenderer` so fenced blocks and inline spans are excluded while the surrounding prose still translates. Confirmed DOMPurify preserves `translate` under `options.sanitize` before relying on this.

### Deliberately left translatable

`TypePrefix` ("Array of"), `TypeTitle`, `RecursiveLabel`, required / optional / deprecated labels, constraints, every `l()` label, `UnderlinedHeader` section headings ("query Parameters", "Response Schema"), sidebar `sidebarLabel` summaries, `info.title`, response summaries, and descriptions.

### Two judgment calls left open

**`TypeName` is not pure code.** `displayType` (`services/models/Schema.ts`) joins type keywords with the English word `' or '`, and in the `oneOf` path embeds `typePrefix` and `schema.title`. Marking `TypeName` blocks translation of those too. I kept it marked, since the OAS type keywords are the code-sensitive part and the actual complaint in #2696, but splitting them properly means restructuring `displayType` into parts — a larger change than this PR should carry. Happy to do it here if you would prefer.

**`Dropdown` options are mixed.** The same component renders media types (code), discriminator schema titles, and example names. I marked `MimeLabel`, which is unambiguously media types, and left `Dropdown` alone rather than guess.

### Verification

279 tests pass across 28 suites; 5 snapshots updated. `tsc --noEmit` reports no errors in `src`. Added `src/services/__tests__/mdtranslate.test.ts` for the markdown renderer, including an assertion that surrounding prose stays unmarked. I inspected the `ConstraintItem` snapshot diff specifically to confirm the only changes there were attribute removals on that element.

displague:
@AlexVarchuk I compared notes (with Claude) against two other PRs in the a11n and i18n space and made some improvements here to get the PR to cover the right elements and to get it more inline with repo expectations. I'm hoping these updates and review comments are in the direction.

PR #2726, opened 2025-09-26
chore: v2.5.1

DmitryAnansky:
## What/Why/How?

 Update mobx-react version

## Check yourself

- [x] Code is linted
- [x] Tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [2.5.1](https://github.com/Redocly/redoc/compare/v2.4.0...v2.5.1) (2025-09-26)
+
+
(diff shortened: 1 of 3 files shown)


AlexVarchuk (approved):

talkinnl:
Why did you remove the `^` ?

AlexVarchuk:
Hi @talkinnl , the reason behind this you could find here https://github.com/mobxjs/mobx/issues/4585.
So to prevent similar problems in future we decide to remove ^

talkinnl:
Thanks for the quick and clear reply!

I think it’s good that the reason is now documented :)

PR #2772, opened 2026-02-09
feat: respect titles for oneOf variants

susliko:
## What/Why/How?
Hi 👋🏻 
This fixes https://github.com/Redocly/redoc/pull/2772#issue-3916359410 when buttons for `oneOf` cases do not pick up schema names (or titles, if provided) and just show "object" instead.

The fix is to avoid adding `default: undefined` property to the FieldModel as it makes further processing fall into [allOf cases](https://github.com/susliko/redoc/blob/c843ce27663457133cd5fd884491139f17fb1e3f/src/services/OpenAPIParser.ts#L151), which leads to further unwanted changes of structure in [hoistOneOfs](https://github.com/susliko/redoc/blob/c843ce27663457133cd5fd884491139f17fb1e3f/src/services/OpenAPIParser.ts#L378)

## Screenshots
Before:
image

After:
image

## Check yourself

- [x] Code is linted
- [x] Tested


jeremyfiel:
@AlexVarchuk    are you still taking improvements on redoc community?

